PR #232, opened 2025-12-10
feat(mcp): Add MCP server routes & dedicated customization

EnkiP:
### General

### Security

### DIFF
--- a/packages/forest_admin_agent/spec/lib/forest_admin_agent/mcp/oauth_provider_spec.rb
+++ b/packages/forest_admin_agent/spec/lib/forest_admin_agent/mcp/oauth_provider_spec.rb
@@ -1,0 +1,329 @@
+require 'spec_helper'
+
+module ForestAdminAgent
+  module Mcp
+    describe OauthProvider do
+      subject(:oauth_provider) { described_class.new(forest_server_url: 'https://api.forestadmin.com') }
+
+      let(:auth_secret) { 'test-auth-secret-key-for-jwt-signing' }
+      let(:env_secret) { 'test-env-secret' }
+      let(:mock_http_client) { instance_double(Faraday::Connection) }
+
+      before do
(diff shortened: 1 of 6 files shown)


EnkiP:
## Code review

Found 2 issues:

1. **Missing require for error classes causes NameError at runtime** - The `OAuthProvider` class raises custom errors (`UnsupportedTokenTypeError`, `InvalidTokenError`, `InvalidClientError`) but never requires `errors.rb`. With Zeitwerk autoloading, these error classes won't be loaded because the file is named `errors.rb` (expecting an `Errors` constant) while the code references individual error class names. This will cause `NameError: uninitialized constant` when any of these errors are raised.

https://github.com/ForestAdmin/agent-ruby/blob/4175a99e85d46c78c29bbdeeb029f1903a91aa84/packages/forest_admin_agent/lib/forest_admin_agent/mcp/oauth_provider.rb#L1-L4

Fix: Add `require_relative 'errors'` at the top of `oauth_provider.rb` and `token_generator.rb`, or split error classes into individual files following Zeitwerk naming conventions.

2. **Open redirect vulnerability in OAuth authorize endpoint** - The `redirect_uri` parameter is used directly in error redirects without validating it against the client's registered redirect URIs. Per OAuth 2.0 (RFC 6749 Section 3.1.2.2), the authorization server MUST validate redirect_uri against pre-registered URIs. An attacker can set `redirect_uri` to a malicious URL, and if client validation fails, the error response will redirect to the attacker's site.

https://github.com/ForestAdmin/agent-ruby/blob/4175a99e85d46c78c29bbdeeb029f1903a91aa84/packages/forest_admin_agent/lib/forest_admin_agent/routes/mcp/oauth_authorize.rb#L33-L35

Fix: Before using `redirect_uri` in `error_redirect`, validate it against `client['redirect_uris']` returned by `get_client()`. If the URI doesn't match any registered URI, return an error response directly instead of redirecting.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #288, opened 2026-04-27
feat(zendesk): add Zendesk datasource package

PMerlet:
## Summary

New `forest_admin_datasource_zendesk` package that surfaces Zendesk Tickets, Users, Organizations, and ticket Comments as Forest Admin collections — same architectural pattern as the existing `forest_admin_datasource_active_record` / `_mongoid` / `_rpc` siblings, mirroring the Node `@forestadmin/datasource-*` ecosystem.

- **4 collections** with schema-level intra-datasource relations: Ticket → requester/assignee/organization/comments, Comment → author/ticket, User → organization/requested_tickets, Organization → users/tickets.
- **Custom fields**: discovers admin-defined `ticket_fields`, `user_fields`, `organization_fields` at boot and exposes them as real Forest columns. Translator rewrites custom-column names to Zendesk Search syntax.
- **Timezone-aware `Date` filters**: Forest's caller timezone is threaded through the translator; bare `Date` values are interpreted as start-of-day in the caller's TZ then converted to UTC (DST-aware).
- **Loud-vs-graceful error handling**: critical paths (`search`, `count`, `fetch_ticket_comments`) raise `APIError`; enrichment paths (bulk user/org lookups, schema introspection) log a warning and degrade safely.
- **Synthetic single-PK on Comments** (`<comment_id>-<ticket_id>` String) to side-step `forest_admin_rails 1.26.2`'s URL constraint, which rejects the `|` character used by the toolkit's native composite-key `pack_id`. Filed-this-as-a-bug-worth-fixing-upstream note in the relevant code comment.

## What's NOT in this PR (deliberate v1 scope)

- Read-only — no `create`/`update`/`delete` on any collection. The Compliance handoff workflow (update ticket status / reassign group) is the obvious next add.
- No caching — every list page hits Zendesk's Search API + bulk user/org lookups. Add `Rails.cache` wrapping in a follow-up if rate limits become a concern.
- Zendesk Search's 1000-result hard cap is not surfaced in pagination yet (cursor-based Incremental Export API would be the correct path for large datasets).
- No native OR-aggregator translation (raises loudly on unsupported operators).

## Test plan

- [x] `cd packages/forest_admin_datasource_zendesk && bundle install && bundle exec rspec` — 131 examples, 0 failures
- [x] Coverage ≥ 90% line and branch (98.5% line / 90.7% branch with WebMock blocking real net)
- [x] Manually verified end-to-end against a live Zendesk dev instance (browse tickets, requester join, conversation thread via Ticket → comments related list, comment detail page via synthetic PK)
- [ ] Reviewer to confirm: gemspec metadata, dependency pins, SchemaEmitter compatibility with toolkit ≥ 1.26
- [ ] Reviewer to confirm: package layout matches conventions (Zeitwerk, Rakefile, .rspec)

> [!NOTE]
> ### Add Zendesk datasource package with collections for tickets, users, organizations, and comments
> - Introduces a new `forest_admin_datasource_zendesk` gem that wraps the Zendesk API as a Forest Admin datasource, supporting list, count, create, update, and delete for tickets, users, organizations, and comments.
> - [`Client`](https://github.com/ForestAdmin/agent-ruby/pull/288/files#diff-fc22bbff66143ab7470df0956b3d3bb434034f976495335db51dff5899505000) wraps `zendesk_api` with consistent error handling: critical paths raise `APIError`, non-critical paths log and return safe defaults.
> - [`ConditionTreeTranslator`](https://github.com/ForestAdmin/agent-ruby/pull/288/files#diff-04874fba270e8c2c372e66e9708ec03872e038967dd7689de92e81177fa1b949) converts Forest condition trees to Zendesk Search query strings, with timezone-aware date handling and custom field mapping.
> - [`CustomFieldsIntrospector`](https://github.com/ForestAdmin/agent-ruby/pull/288/files#diff-2ec7308f0fb6cb7fbde7cc62cacaa8a690d33f647891e28715ee777474c03500) fetches Zendesk custom field definitions at boot and produces typed `ColumnSchema` entries for tickets, users, and organizations.
> - The `Comment` collection is read-only and only returns results when a `ticket_id` scope is present in the filter; it uses a synthetic composite `<comment_id>-<ticket_id>` primary key.
> - Release pipeline in [`.releaserc.js`](https://github.com/ForestAdmin/agent-ruby/pull/288/files#diff-3226d814d68c694678e5341c4d26cd2f86c642124ca5cf683511de900a2dcd23) and CI in [`.github/workflows/build.yml`](https://github.com/ForestAdmin/agent-ruby/pull/288/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721) are extended to build, publish, and report coverage for the new gem.
>
> 
> #### Changes since #288 opened
>
> - Refactored `collections.Ticket` to use four new mixin modules and changed comments from a `OneToMany` relation to a read-only structured array column [22d782e]
> - Modified `ForestAdminDatasourceZendesk::Query::ConditionTreeTranslator.format_string` to always quote and escape string filter values [22d782e]
> - Refactored `ForestAdminDatasourceZendesk::Client` to extract write and introspection operations into separate modules and added bulk ticket fetching [22d782e]
> - Centralized Zendesk query string building into `collections.BaseCollection.build_zendesk_query` and refactored dependent methods to use it [22d782e]
> - Removed `Comment` collection registration from `ForestAdminDatasourceZendesk::Datasource` [22d782e]
> - Refactored `collections.BaseCollection.project` to use `to_h` for projection filtering [22d782e]
> - Updated test coverage for `ForestAdminDatasourceZendesk::Query::ConditionTreeTranslator` value quoting [22d782e]
> - Updated `packages/forest_admin_datasource_zendesk/collections/ticket_spec.rb` to reflect schema and behavior changes [22d782e]
> - Added `packages/forest_admin_datasource_zendesk/forest_admin_datasource_zendesk.gemspec` to Gemspec/RequireMFA exclusions and set metadata `rubygems_mfa_required` to false [22d782e]
> - Removed per-package RuboCop configuration and cleaned up documentation comments [22d782e]
> - Wrapped RSpec specs in `ForestAdminDatasourceZendesk` module with included `ForestAdminDatasourceToolkit::Components::Query` and locally-rebound constants [ed28d88]
> - Refactored RSpec expectations from pre-invocation `expect(...).to receive` to post-invocation `allow(...).to receive` with `expect(...).to have_received` assertions [ed28d88]
> - Added enhanced assertions in `Collections::Ticket` spec for client interaction verification [ed28d88]
> - Introduced `leaf_class` and `branch_class` helpers in `Query::ConditionTreeTranslator` spec [ed28d88]
> - Added `packages/forest_admin_datasource_zendesk/spec/**/*` to `Metrics/ModuleLength` exclusion list in `.rubocop.yml` [ed28d88]
> - Reformatted test setup code for readability in `Client` and `Schema::CustomFieldsIntrospector` specs [ed28d88]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4ac39d5.</sup>
> 
>
>


packages/forest_admin_datasource_zendesk/.rubocop.yml
matthv, line 1: This file should be removed: no other package has its own `.rubocop.yml`, they all inherit from the root config. And the overrides here actually loosen the limits (Metrics/ClassLength: 230, Metrics/MethodLength: 25, Layout/LineLength: 130) rather than respecting the project's

packages/forest_admin_datasource_zendesk/Rakefile
matthv, line 1: Looks like the Rakefile here isn't actually used (CI and release both bypass rake). Worth dropping?

packages/forest_admin_datasource_zendesk/forest_admin_datasource_zendesk.gemspec
```diff
@@ -0,0 +1,35 @@
...
+  spec.homepage    = 'https://www.forestadmin.com'
+  spec.summary     = 'Zendesk datasource for Forest Admin Ruby agent.'
+  spec.description = 'Surface Zendesk tickets, users, organizations and comments as Forest Admin collections.'
+  spec.license     = 'GPL-3.0'
+  spec.required_ruby_version = '>= 3.0.0'
+
+  spec.metadata['homepage_uri']    = spec.homepage
+  spec.metadata['source_code_uri'] = 'https://github.com/ForestAdmin/agent-ruby'
+  spec.metadata['changelog_uri']   = 'https://github.com/ForestAdmin/agent-ruby/blob/main/CHANGELOG.md'
+  spec.metadata['rubygems_mfa_required'] = 'true'
```
matthv: 'rubygems_mfa_required' => 'true' diverges from the other gemspecs (all set to 'false' and listed under Gemspec/RequireMFA: Exclude in the root rubocop). Security-wise it's better, but it deserves a monorepo-wide decision rather than a one-off opt-in on a single package

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/collections/searchable.rb
```diff
@@ -0,0 +1,60 @@
...
+                                 query: compose_full_query(caller, filter),
+                                 sort_by: sort_by, sort_order: sort_order,
+                                 page: page, per_page: per_page)
+      end
+
+      # Both list and count must build the query the same way: condition tree
+      # AND `filter.search`. A previous version of search_records omitted the
+      # search term, which made the count badge disagree with the rendered
+      # list ("100 results, count says 5") — guard against that by sharing
+      # this builder.
```
matthv: Since this is a new datasource, the "previous version" referenced here only ever existed during this PR's development. Once merged, the comment loses its anchor

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/collections/base_collection.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      # Sort entries arrive either as Sort::Clause objects (responding to
+      # `field`/`ascending`) or as plain hashes (the toolkit normalises them
+      # at construction time, but specs and a few code paths still build them
+      # by hand). Handle both.
+      #
+      # `key?` (rather than `||`) for the boolean: `entry[:ascending] || ...`
+      # would silently flip a descending sort to ascending if both symbol and
+      # string keys exist with different values, and falls through to the
+      # other key whenever ascending is explicitly false.
```
matthv: The invariant is already covered by `ticket_spec.rb:130` (preserves ascending=false even when both symbol and string keys are present). The test name says enough. This comment, and the duplicated # Regression: block in the test itself, can both be dropped.

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/query/condition_tree_translator.rb
```diff
@@ -0,0 +1,143 @@
...
+    # Custom-field translation: callers pass `custom_fields:` (a hash from
+    # Forest column names to Zendesk Search field names, owned by the
+    # Datasource instance) so multi-tenant agents with several Zendesk
+    # datasources don't trample each other's mappings.
+    #
+    # Timezone handling: callers pass `timezone:`; Date values are
+    # interpreted as start-of-day in that TZ, then converted to UTC.
+    # Time/DateTime values are converted to UTC directly (they already carry
+    # offset info). String values are passed through verbatim, with internal
+    # double quotes escaped when wrapping in quotes is needed.
```
matthv: Class-level docs are fine in Ruby, but this preamble mixes contract with implementation paraphrase. The operator list will rot the first time we add one (the case in translate_leaf is the source of truth), and the Time/DateTime/string-quoting details are implementation, not contract. 
Suggested trim:
```
# Translates a Forest ConditionTree into a Zendesk Search API query string.
# See https://developer.zendesk.com/api-reference/ticketing/ticket-management/search/
#
# Contract:
# - Unsupported operators raise UnsupportedOperatorError (no silent fallback).
# - `custom_fields` is owned by the Datasource instance so multi-tenant
#   agents don't share state across Zendesk subdomains.
# - Date values are interpreted as start-of-day in `timezone`, then UTC.
```

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/client.rb
```diff
@@ -0,0 +1,206 @@
...
+
+    def fetch_organization_fields
+      best_effort('fetch_organization_fields (custom fields will be unavailable)', default: []) do
+        Array(api.connection.get('organization_fields').body['organization_fields'])
+      end
+    end
+
+    def raw
+      api
+    end
```
matthv: No callers for this method anywhere in the PR. Looks like dead code ?

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/collections/ticket.rb
```diff
@@ -0,0 +1,272 @@
...
+          else
+            h[k] = v
+          end
+        end
+        [custom, rest]
+      end
+
+      def fetch_records(filter, timezone)
+        ids = extract_id_lookup(filter.condition_tree)
+        return ids.filter_map { |id| datasource.client.find_ticket(id) } if ids
```
matthv: This line triggers one GET /tickets/:id call per id. With id IN [1..50], that's 50 sequential HTTP round-trips (multiple seconds of latency).

Zendesk has a `/tickets/show_many?ids=…` endpoint that returns up to 100 tickets in a single call. The bulk_show_many helper in `client.rb `already implements that pattern for users and organizations (see fetch_users_by_ids / fetch_organizations_by_ids). Worth applying the same approach to tickets to stay consistent and avoid the latency cliff.

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/collections/comment.rb
matthv, line 1: Reconsidering the Comment collection design
 
The log + return [] when list is called without a ticket scope (lines 19-27) is the visible symptom of a deeper layering issue. Walking through the current state:
 
 1. `Comment` is registered as a top-level collection, so it appears in the Forest sidebar alongside Ticket / User / Organization.
 2. When a user clicks it, Forest calls list(caller, filter, projection) with no ticket_id in the filter.
3. The collection logs an info line nobody will ever read, `returns []`, and the user sees an empty grid with no explanation.
4. To get a stable URL for an individual comment via Ticket > comments, the schema invents a synthetic <comment_id>-<ticket_id> String PK to dodge forest_admin_rails 1.26.2's URL constraint.

That's a lot of complexity (sidebar entry, scope resolution, synthetic PK, custom extract_field_lookup, Branch-walking logic) to support a use case Zendesk's API doesn't actually expose: standalone comment browsing. There's no /comments/{id} endpoint, comments are always fetched as /tickets/{id}/comments.

Why "just hide the sidebar entry" doesn't work
Checked the publication decorator: remove_collection cascades. published? in PublicationCollectionDecorator:79-85 will unpublish `Ticket > comments` as soon as `Comment` is removed. The toolkit doesn't currently expose a "hide from sidebar but keep accessible via relation" flag at the collection level.

Proposed simpler design: drop the Comment collection

Comments are read-only here (no create/update/delete/aggregate). The only valid access pattern is "give me the comments of this ticket." An embedded structured field on Ticket models that cleanly.

add_external_relation in the customizer layer does this, but it's the wrong layer for a datasource gem. We can do the same thing natively at the datasource layer because ColumnSchema#column_type accepts a structured [hash] shape and Ticket#list already controls projection-aware fetching.

```
# In Ticket#initialize:
COMMENT_THREAD_SCHEMA = {
 'id'           => 'Number',
'body'         => 'String',
'html_body'    => 'String',
'public'       => 'Boolean',
'author_email' => 'String',
'author_name'  => 'String',
'created_at'   => 'Date'
}.freeze
```

```
add_field('comments', ColumnSchema.new(
  column_type: [COMMENT_THREAD_SCHEMA],
  filter_operators: []
))
```

```
# In Ticket#list:
def list(caller, filter, projection)
   records = fetch_records(filter, timezone_for(caller))
   rows = build_rows(records, projection, caller)
   rows = embed_comments(records, rows) if want_comments?(projection)
   rows
end

def want_comments?(projection)
  projection.nil? || Array(projection).map(&:to_s).any? { |p| p == 'comments' || p.start_with?('comments:') }
end
```

The lazy gating mirrors the existing needs_requester_email? pattern, so list views (where projection is always explicit) don't trigger N+1.

What disappears

- comment.rb (the entire Comment collection)
- The synthetic <comment_id>-<ticket_id> PK and the forest_admin_rails 1.26.2 workaround comment
- extract_field_lookup, the Branch traversal, decoded_synthetic_pairs, decode_synthetic_id
- The OneToManySchema comments relation on Ticket (replaced by the structured field)
- The log + return [] block this comment thread is anchored on
- comment_spec.rb in its entirety, replaced by smaller projection-gated tests on ticket_spec.rb

Trade-off
 author_id becomes a flat author_email / author_name denormalized in the embedded structure, instead of a clickable FK to ZendeskUser. For a v1 read-only thread view it's acceptable: users read the conversation on the ticket page and don't typically jump to the agent's profile from a comment. If that need surfaces, we add it in a follow-up by reintroducing the relation.

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/query/condition_tree_translator.rb
```diff
@@ -0,0 +1,143 @@
...
+        )
+        value.strftime('%Y-%m-%dT00:00:00Z')
+      end
+
+      # Strings with whitespace OR internal double quotes need quoting so
+      # Zendesk parses them as a single phrase. We backslash-escape internal
+      # quotes per Zendesk's documented quoting rules; without this, a value
+      # like `test "with" quotes` would emit a malformed query.
+      def format_string(value)
+        return value unless value.match?(/[\s"]/)
```
matthv: The regex /[\s"]/ only triggers quoting on whitespace or double quotes. Zendesk Search has other characters that change query semantics: ( ) for grouping, : for field separators, - for negation. A value like (test gets emitted bare and Zendesk parses subject:(test as a grouping expression instead of a literal match.

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/collections/searchable.rb
```diff
@@ -0,0 +1,60 @@
...
+      def compose_full_query(caller, filter)
+        translated = ForestAdminDatasourceZendesk::Query::ConditionTreeTranslator.call(
+          filter.condition_tree, timezone: timezone_for(caller),
+                                 custom_fields: datasource.custom_field_mapping
+        )
+        [translated, filter.search].compact.reject(&:empty?).join(' ')
+      end
```
matthv: Same query builder lives in two places: compose_full_query(caller, filter) here and Ticket#build_query(filter, timezone) in ticket.rb:119. Bodies are identical apart from how they get the timezone (mixin reads timezone_for(caller) from BaseCollection, Ticket receives it as an arg).

 Worth lifting into BaseCollection#build_zendesk_query(caller, filter) and having both Searchable and Ticket call it. Single source of truth for the translator + filter.search composition, and one less place for them to drift apart.

Sketch:

```
# base_collection.rb
def build_zendesk_query(caller, filter)
   translated = ForestAdminDatasourceZendesk::Query::ConditionTreeTranslator.call(
    filter.condition_tree,
    timezone: timezone_for(caller),
    custom_fields: datasource.custom_field_mapping
  )
 [translated, filter.search].compact.reject(&:empty?).join(' ')
 end

# searchable.rb
def compose_full_query(caller, filter)
   build_zendesk_query(caller, filter)
 end

# ticket.rb (drop the local build_query, update its callers)
def aggregate_count(caller, filter)
  datasource.client.count('ticket', query: build_zendesk_query(caller, filter))
end

def fetch_records(filter, caller)  # signature update: pass caller, drop timezone
  ids = extract_id_lookup(filter.condition_tree)
  return ids.filter_map { |id| datasource.client.find_ticket(id) } if ids

  query = build_zendesk_query(caller, filter)
  ...
end
```

Once factored, compose_full_query is just a thin alias that the mixin can keep for clarity, or Searchable can call build_zendesk_query directly and the alias goes away.

matthv (commented):
The comment density across this package feels much higher than the rest of the monorepo (compare with active_record DS which is nearly comment-free).
A  lot just restate the code. The # Error policy: block at the top doesn't add anything must_succeed / best_effort don't already say.

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/collections/base_collection.rb
```diff
@@ -0,0 +1,118 @@
...
+      end
+
+      # Translates a Forest Page (offset/limit) into Zendesk's [page, per_page].
+      def translate_page(page)
+        return [1, Client::MAX_PER_PAGE] if page.nil?
+
+        per_page = page.limit&.positive? ? [page.limit, Client::MAX_PER_PAGE].min : Client::MAX_PER_PAGE
+        page_num = (page.offset.to_i / per_page) + 1
+        [page_num, per_page]
+      end
```
matthv, line 67: Zendesk Search caps total results at 1000 (in addition to the per-page max of 100, which is already handled via `Client::MAX_PER_PAGE`). The total cap isn't represented anywhere in the code: translate_page happily computes page_num = 11 past offset 1000, sends it to Zendesk, gets a 422 back, and the user sees a generic APIError with no actionable message.
Suggested fix: add a MAX_TOTAL_RESULTS = 1000 constant, clamp page_num to MAX_TOTAL_RESULTS / per_page, and raise a ForestException with a clear message when exceeded:

`MAX_TOTAL_RESULTS = 1000`

packages/forest_admin_datasource_zendesk/.rubocop.yml
christophebrun-forest: Deleted

packages/forest_admin_datasource_zendesk/forest_admin_datasource_zendesk.gemspec
christophebrun-forest: Update and Add the exclude in the root rubocop

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/client.rb
christophebrun-forest: Yes, dead code.

matthv (approved):

PR #291, opened 2026-04-28
fix(polymorphic): skip pre-delete FK nullify when datasource cascades

### LINKED ISSUE #289: `delete_records` nullifies `PolymorphicOneToMany` FKs before destroy, failing when FK is `NOT NULL`
## Description

We've been following the great work on polymorphic association fixes (related to #205 and #279) and wanted to flag another edge case we've run into.

We have a polymorphic association between models for "achievement evidence".

Simplified data structure:
`Feedback → AchievementEvidence`

`AchievementEvidence`:
- `evidence_id` and `evidence_type` stores the polymorphic parent (e.g. `Feedback`)
- `achievement_progress_id` for the progress record it is attached to

## Environment
- `forest_admin_rails` 1.26.3
- Rails 7.x, Ruby 3.x, PostgreSQL

## Reproduction

Schema:
```ruby
create_table "achievement_evidences" do |t|
  t.string "evidence_type", null: false
  t.bigint "evidence_id",   null: false
  t.bigint "achievement_progress_id", null: false
end
```

Models:
```ruby
class Feedback < ApplicationRecord
  has_many :achievement_evidences, as: :evidence, dependent: :destroy
end

class AchievementEvidence < ApplicationRecord
  belongs_to :evidence, polymorphic: true
end
```

## Expected

Deleting a `Feedback` record via the ForestAdmin UI should destroy associated `AchievementEvidence` records. From the Rails console this works correctly:

```ruby
Feedback.find(1).destroy
# DELETE FROM "achievement_evidences" WHERE "achievement_evidences"."evidence_type" = 'Feedback' AND "achievement_evidences"."evidence_id" = 1
# DELETE FROM "feedbacks" WHERE "feedbacks"."id" = 1
```

## Actual

Deleting via the ForestAdmin UI fails with:

```
An error occurred when trying to delete feedback: Your data seems to be invalid:
Validation failed: Evidence is required, Evidence type is not included in the list.
```

From the logs, ForestAdmin is issuing an `UPDATE` to nullify the FK before calling destroy:

```sql
UPDATE "achievement_evidences" SET "evidence_id" = NULL, "evidence_type" = NULL
WHERE "achievement_evidences"."evidence_type" = 'Feedback' AND "achievement_evidences"."evidence_id" = 1
```

Since `evidence_id` is `NOT NULL` in our schema, this update fails validation before the destroy ever runs.

Happy to provide any additional details or help test a fix. Thanks for your time!

bexchauveto:
## Summary

Fixes #289. The agent's `delete_records` action pre-emptively `UPDATE`s polymorphic child rows to `NULL` both the FK id and type column **before** issuing the parent delete. For ActiveRecord parents that declare `dependent: :destroy`, this breaks deletion entirely:

- Rails 5+ `belongs_to :foo, polymorphic: true` adds a presence validator → the UPDATE fails with `Validation failed: <relation> must exist`.
- A `null: false` FK column → the UPDATE violates the NOT NULL constraint.
- Even when the FK is nullable, the manual nullify defeats `dependent: :destroy`: by the time Rails' destroy callback runs, its `WHERE notable_id = ? AND notable_type = ?` matches zero rows, silently orphaning the children.

The reporter showed the bug works correctly from the Rails console (Rails handles the cascade), but fails through Forest because of this pre-delete UPDATE.

## Fix

Plumb a `cascade_on_delete` flag through to the polymorphic schemas; when true, the agent skips its manual nullify and lets the datasource handle the cascade itself.

- **Toolkit** — added `cascade_on_delete` (default `false`) to `PolymorphicOneToManySchema` and `PolymorphicOneToOneSchema`. The flag rides through the RPC layer for free (`**schema` reconstruction + `.to_json` serializer dumps all instance vars).
- **AR datasource** — sets `cascade_on_delete: true` when the association declares `dependent: :destroy`, `:destroy_async`, or `:delete_all`.
- **Agent route** — `delete_records` skips the polymorphic-cleanup loop body when `field_schema.cascade_on_delete` is truthy. Guarded with `respond_to?` for compat with anything the customizer might inject.

Mongoid / Snowflake / RPC keep the existing behavior (`cascade_on_delete` defaults to false, so the manual cleanup still runs).

## Test plan

- [x] `forest_admin_datasource_toolkit` rspec — 463 examples, 0 failures (added defaults + explicit-flag specs on both schemas)
- [x] `forest_admin_datasource_active_record` rspec — 147 examples, 0 failures (added: `cascade_on_delete=true` for `Project.has_many :members, as: :memberable, dependent: :destroy`; `=false` for `User.has_one :address, as: :addressable`)
- [x] `forest_admin_agent` rspec — 683 examples, 0 failures (new `with polymorphic relation that cascades on delete` context: foreign collection's `update` is NOT called, parent `delete` still is)
- [x] `forest_admin_datasource_rpc` rspec — 160 examples, 0 failures (sanity: schema round-trips through RPC unchanged)
- [x] End-to-end repro against a Rails 8.1 / PG 15 app: a `Customer` with `has_many :notes, as: :notable, dependent: :destroy` and `notable_id`/`notable_type` `null: false` — delete now performs the same SQL as `customer.destroy!` (children deleted via cascade, then parent), instead of erroring on the nullify.

> [!NOTE]
> ### Skip FK nullification before delete for polymorphic relations when datasource cascades
> - Adds a `cascade_on_delete` flag (default `false`) to `PolymorphicOneToManySchema` and `PolymorphicOneToOneSchema` in the toolkit.
> - The ActiveRecord datasource sets this flag based on the association's `dependent` option (`:destroy`, `:destroy_async`, or `:delete_all`) via a new `cascade_on_delete?` helper in [`Collection`](https://github.com/ForestAdmin/agent-ruby/pull/291/files#diff-394937cf0ee68635c26a0af527b70b25e67ac962f1137a31cb25f490e36c686f).
> - The delete route in [`Delete#delete_records`](https://github.com/ForestAdmin/agent-ruby/pull/291/files#diff-146db32b6d7a7192178c2bdb7ddedb09860d103a5985db4b0bcafabd99d21114) now skips the pre-delete FK nullify step for polymorphic relations when `cascade_on_delete` is `true`, avoiding redundant or conflicting updates before the DB cascade fires.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 198e837.</sup>
> 
>

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -237,6 +237,8 @@ Metrics/ParameterLists:
     - 'packages/forest_admin_datasource_customizer/lib/forest_admin_datasource_customizer/dsl/builders/form_builder.rb'
     - 'packages/forest_admin_datasource_mongoid/lib/forest_admin_datasource_mongoid/datasource.rb'
     - 'packages/forest_admin_datasource_toolkit/lib/forest_admin_datasource_toolkit/schema/relations/many_to_many_schema.rb'
+    - 'packages/forest_admin_datasource_toolkit/lib/forest_admin_datasource_toolkit/schema/relations/polymorphic_one_to_many_schema.rb'
+    - 'packages/forest_admin_datasource_toolkit/lib/forest_admin_datasource_toolkit/schema/relations/polymorphic_one_to_one_schema.rb'
     - 'packages/forest_admin_datasource_toolkit/lib/forest_admin_datasource_toolkit/schema/column_schema.rb'
     - 'packages/forest_admin_datasource_toolkit/lib/forest_admin_datasource_toolkit/components/caller.rb'
     - 'packages/forest_admin_datasource_toolkit/lib/forest_admin_datasource_toolkit/components/query/filter.rb'
--- a/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/delete.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/delete.rb
@@ -41,6 +41,7 @@ module ForestAdminAgent
 
           context.collection.schema[:fields].each_value do |field_schema|
             next unless ['PolymorphicOneToOne', 'PolymorphicOneToMany'].include?(field_schema.type)
+            next if field_schema.respond_to?(:cascade_on_delete) && field_schema.cascade_on_delete
 
             origin_values = selection_ids[:ids].map do |pk_hash|
               if pk_hash.is_a?(Hash)
--- a/packages/forest_admin_agent/spec/lib/forest_admin_agent/routes/resources/delete_spec.rb
+++ b/packages/forest_admin_agent/spec/lib/forest_admin_agent/routes/resources/delete_spec.rb
@@ -380,6 +380,105 @@ module ForestAdminAgent
               end
             end
           end
+
+          context 'with polymorphic relation that cascades on delete' do
+            before do
+              cache = FileCache.new('app', 'tmp/cache/forest_admin')
+              cache.clear
+
+              agent_factory = ForestAdminAgent::Builder::AgentFactory.instance
+              agent_factory.setup(
+                {
+                  auth_secret: 'cba803d01a4d43b55010cab41fa1ea1f1f51a95e',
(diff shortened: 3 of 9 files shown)


matthv (approved):
LGTM

PR #292, opened 2026-04-28
feat(snowflake): add read-only Snowflake datasource gem

bexchauveto:
## Summary

Introduces `forest_admin_datasource_snowflake` — a native Forest Admin datasource backed by Snowflake via ODBC, with no ActiveRecord involvement.

Read-only first pass. Covers list/aggregate query translation, bulk schema introspection, connection pooling, retry-on-disconnect, foreign-key auto-discovery, primary-key resolution (incl. composite keys with a strict operator override), and optional per-session statement timeout. Wired into `build.yml` lint/test matrices with a `unixodbc-dev` apt step and into `.releaserc.js` for the release pipeline.

## Highlights

**Schema introspection**
- Each datasource instance is locked to a single Snowflake schema. `Schema=` from the connection string wins; otherwise the datasource snapshots `CURRENT_SCHEMA()` once at boot. If neither yields a value, boot raises with a clear "set Schema=<name>" message instead of letting cross-schema collisions happen.
- A single bulk `INFORMATION_SCHEMA.COLUMNS` query introspects every column in the schema in one round-trip. Combined with `conn.tables`, `SHOW PRIMARY KEYS IN SCHEMA`, and `SHOW IMPORTED KEYS IN SCHEMA`, total introspection cost is **4 queries regardless of table count** (was ~2N+3). Observed agent boot dropped from ~63s to ~5s on a 7-table schema.
- Primary key resolution: operator `primary_keys:` override → Snowflake-declared PK (composite preserved, ordered by `key_sequence`) → column literally named `id` → first column. Override accepts a `String` or `Array<String>` for composite keys; a typo'd column name raises `ForestAdminDatasourceSnowflake::Error` at boot rather than silently falling back.
- FK auto-discovery via `SHOW IMPORTED KEYS IN SCHEMA` runs unconditionally; `discover_relations` exposes each FK as a `ManyToOne` relation. Cross-datasource FKs (Snowflake → Postgres) still need to be wired manually with `add_many_to_one_relation` / `add_one_to_many_relation` — Snowflake has no concept of an FK to another database.
- Introspection failures (column / PK / FK) emit a `[forest_admin_datasource_snowflake]` prefixed warning to stderr and cache the empty result so the same broken query isn't re-issued on every collection lookup.

**Type translation**
- Forest types are derived from the Snowflake-native `DATA_TYPE` returned by `INFORMATION_SCHEMA.COLUMNS`. The mapping covers the full canonical Snowflake set — `NUMBER`/`DECIMAL`/`INT`/`BIGINT`/`SMALLINT`/`TINYINT`/`BYTEINT`, `FLOAT`/`FLOAT4`/`FLOAT8`/`DOUBLE`/`REAL`, `VARCHAR`/`CHAR`/`TEXT`/`STRING`, `BOOLEAN`, `DATE`, `TIME`, `DATETIME`/`TIMESTAMP`/`TIMESTAMP_NTZ`/`LTZ`/`TZ`, `VARIANT`/`OBJECT`/`ARRAY`, `BINARY`/`VARBINARY`, `GEOGRAPHY`/`GEOMETRY`/`VECTOR`. Unknown types fall back to `String`.
- `ODBC::Date` / `ODBC::TimeStamp` / `ODBC::Time` coerced to native Ruby `Date`/`Time`/string (otherwise they JSON-serialize as `{}`).
- `VARIANT` / `OBJECT` / `ARRAY` columns are JSON-parsed at projection time so the UI receives structured data.
- Session-level `TIMEZONE='UTC'` so all three TIMESTAMP variants serialize consistently.

**SQL translation correctness**
- `EQUAL` / `NOT_EQUAL` with `nil` value emit `IS NULL` / `IS NOT NULL` (used to bind NULL into `field = ?` and silently match nothing under three-valued logic).
- `IN` / `NOT_IN` with `nil` in the list split into `(field IS NULL OR field IN (...))` and `(field IS NOT NULL AND field NOT IN (...))` (used to silently drop or zero-out matches).
- `IN` / `NOT_IN` with an empty list emit `1=0` / `1=1`.
- `COUNT` with a blank/missing field falls through to `COUNT(*)`; `SUM` / `AVG` / `MIN` / `MAX` raise rather than emitting `SUM()`.
- Projection containing only relation paths (`customer:name`) defaults to `SELECT *` instead of `SELECT  FROM ...`.
- Malformed connection-string options (e.g. `Driver=Snowflake;SomeFlag;Server=X`) raise a clear `Malformed connection string option 'SomeFlag': expected 'key=value'` instead of `ArgumentError: wrong array length`.

**Authentication**
- The datasource doesn't interpret connection-string options. Anything the Snowflake ODBC driver accepts works pass-through, including the production-friendly key-pair / JWT flow (`AUTHENTICATOR=SNOWFLAKE_JWT;PRIV_KEY_FILE=/path/to/key.p8`), `EXTERNALBROWSER` SSO, and `OAUTH`.

**Operational**
- `connection_pool`-backed pool, sized via `pool_size:` (default 5) and `pool_timeout:` (default 5s).
- `with_connection` retries the block once on connection-lost ODBC errors and cycles the pool to drop stale handles. Retry patterns include Snowflake auth token expiry (`390114`).
- Optional `statement_timeout:` (in seconds) issues `ALTER SESSION SET STATEMENT_TIMEOUT_IN_SECONDS` on each new connection.
- All ODBC statements are closed via `begin/ensure` on every introspection and execution path.

**Read-only enforcement**
- Every column is emitted with `is_read_only: true`, so Forest's schema emitter computes collection-level `isReadOnly: true` automatically and the UI hides create/edit/delete affordances.
- `create` / `update` / `delete` raise `ForestException` with an explicit read-only message as a defence-in-depth guard.

## Coverage

110 specs across:
- `Parser::Column` — Snowflake-native type mapping, operator sets per Forest type
- `Utils::Identifier` — case-preserving quoting
- `Utils::Query` — condition tree translation, IN/LIKE/ILIKE, sort/page, aggregation, nil semantics
- `Datasource` — introspection, pool, session settings, retry, FK discovery, schema resolution, primary-key resolution, malformed conn_str
- `Collection` — schema introspection, list/aggregate, JSON parsing, write guards

`rubocop` is pinned (`1.86.1`, `rubocop-performance 1.26.1`, `rubocop-rspec 3.9.0`) in both `Gemfile` and `Gemfile-test` so contributors and CI run the same cops.

## Caveats / follow-ups

- **No real-Snowflake integration test in CI** — coverage is fully mocked at the ODBC boundary. A follow-up could add an env-gated smoke test (skipped unless `SNOWFLAKE_*` env vars are present).
- **Read-only by design** for v1. Write support is a future direction; the read-only guards in `Collection` are explicit so a later subclass can override safely.
- **Cross-datasource FKs** (e.g. Snowflake `BILLING_USAGE.CUSTOMER_ID` → Postgres `customers.id`) cannot be auto-discovered — Snowflake only knows about its own metadata.
- **One schema per datasource instance.** To expose tables from multiple Snowflake schemas, instantiate one datasource per schema.

## Test plan

- [ ] `cd packages/forest_admin_datasource_snowflake && BUNDLE_GEMFILE=Gemfile-test bundle install && BUNDLE_GEMFILE=Gemfile-test bundle exec rspec` passes (110 examples, 0 failures)
- [ ] `bundle exec rubocop packages/forest_admin_datasource_snowflake` passes (clean)
- [ ] `bundle exec rubocop` repo-wide passes (still clean after `.rubocop.yml` exclusion additions)
- [ ] Live Forest UI smoke test against a Snowflake account: list, count, filter, aggregate, FK auto-discovery on intra-Snowflake relations, type rendering for VARIANT/OBJECT/ARRAY/BINARY/TIMESTAMP variants, composite-PK record URLs

> [!NOTE]
> ### Add read-only Snowflake datasource gem backed by ODBC
> - Introduces a new `forest_admin_datasource_snowflake` gem that connects to Snowflake via ODBC and exposes tables as read-only Forest Admin collections.
> - `Datasource` manages a connection pool, introspects `INFORMATION_SCHEMA.COLUMNS` and `SHOW PRIMARY KEYS` for schema discovery, and auto-discovers `ManyToOne` relations via `SHOW IMPORTED KEYS`.
> - `Collection` maps Snowflake types to Forest Admin field types, coerces ODBC temporal values to Ruby types, and parses `VARIANT` columns as JSON; `create`, `update`, and `delete` always raise a read-only error.
> - `Utils::Query` translates Forest Admin condition trees into parameterized SQL with support for `COUNT`/`SUM`/`AVG`/`MIN`/`MAX` aggregations, `ORDER BY`, and `LIMIT`/`OFFSET`.
> - Transient connection loss triggers a one-time automatic pool reset and retry; session is always initialized with `TIMEZONE=UTC` and optional schema/statement-timeout settings.
> - CI, release pipeline, and RuboCop config are updated to include the new package.
>
> 
> #### Changes since #292 opened
>
> - Added automatic schema resolution to `ForestAdminDatasourceSnowflake::Datasource` when Schema parameter is not provided in connection string [a305194]
> - Fixed COUNT aggregation handling for empty field values in `ForestAdminDatasourceSnowflake::Utils::Query` [a305194]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1c4a27f.</sup>
> 
>
>

### DIFF
--- a/packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/collection.rb
+++ b/packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/collection.rb
@@ -82,6 +82,9 @@ module ForestAdminDatasourceSnowflake
 
     def resolve_primary_keys(rows)
       column_names = rows.map { |r| r[3] }
+      override = @datasource.primary_keys_override_for(@table_name)
+      return resolve_override_primary_keys(override, column_names) if override
+
       declared_pks = @datasource.primary_keys_for(@table_name)
       matches = declared_pks.filter_map do |pk|
         column_names.find { |name| name.to_s.casecmp(pk.to_s).zero? }
@@ -90,6 +93,15 @@ module ForestAdminDatasourceSnowflake
 
       fallback = (rows.find { |r| r[3].to_s.casecmp('id').zero? } || rows.first)&.dig(3)
       fallback ? [fallback] : []
+    end
+
+    def resolve_override_primary_keys(override, column_names)
+      override.map do |declared_pk|
+        column_names.find { |name| name.to_s.casecmp(declared_pk.to_s).zero? } ||
+          raise(ForestAdminDatasourceSnowflake::Error,
+                "primary_keys override '#{declared_pk}' does not match any column on table " \
+                "'#{@table_name}' (available: #{column_names.join(", ")})")
+      end
     end
 
     def execute_to_hashes(sql, binds, projected_columns)
--- a/packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/collection.rb
+++ b/packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/collection.rb
@@ -45,25 +45,15 @@ module ForestAdminDatasourceSnowflake
     private
 
     def fetch_fields
-      rows = @datasource.with_connection do |conn|
-        stmt = conn.columns(@table_name)
-        begin
-          stmt.fetch_all || []
-        ensure
-          stmt.drop
-        end
-      end
-
-      native_types = @datasource.fetch_snowflake_native_types(@table_name)
+      rows = @datasource.snowflake_columns_for(@table_name)
       pk_names = resolve_primary_keys(rows)
 
       rows.each do |row|
-        column_name = row[3]
-        odbc_type   = row[4]
-        nullable    = row[10] != 0
+        column_name    = row[1]
+        snowflake_type = row[2]
+        nullable       = row[3].to_s.casecmp('YES').zero?
 
-        forest_type = Parser::Column.forest_type_for_snowflake_native(native_types[column_name]) ||
-                      Parser::Column.forest_type_for(odbc_type)
+        forest_type = Parser::Column.forest_type_for_snowflake_native(snowflake_type)
         @json_columns << column_name if forest_type == 'Json'
 
         field = ForestAdminDatasourceToolkit::Schema::ColumnSchema.new(
@@ -81,7 +71,7 @@ module ForestAdminDatasourceSnowflake
     end
 
     def resolve_primary_keys(rows)
-      column_names = rows.map { |r| r[3] }
+      column_names = rows.map { |r| r[1] }
       override = @datasource.primary_keys_override_for(@table_name)
       return resolve_override_primary_keys(override, column_names) if override
 
@@ -91,7 +81,7 @@ module ForestAdminDatasourceSnowflake
       end
       return matches if matches.any?
 
-      fallback = (rows.find { |r| r[3].to_s.casecmp('id').zero? } || rows.first)&.dig(3)
+      fallback = (rows.find { |r| r[1].to_s.casecmp('id').zero? } || rows.first)&.dig(1)
       fallback ? [fallback] : []
     end
 
--- a/packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/collection.rb
+++ b/packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/collection.rb
@@ -56,7 +56,7 @@ module ForestAdminDatasourceSnowflake
       end
 
       native_types = @datasource.fetch_snowflake_native_types(@table_name)
-      pk_name = (rows.find { |r| r[3].to_s.casecmp('id').zero? } || rows.first)&.dig(3)
+      pk_name = resolve_primary_key(rows)
 
       rows.each do |row|
         column_name = row[3]
@@ -80,6 +80,17 @@ module ForestAdminDatasourceSnowflake
         @primary_keys << column_name if field.is_primary_key
         add_field(column_name, field)
       end
+    end
+
+    def resolve_primary_key(rows)
+      column_names = rows.map { |r| r[3] }
+      declared_pk = @datasource.primary_key_for(@table_name)
+      if declared_pk
+        match = column_names.find { |name| name.to_s.casecmp(declared_pk.to_s).zero? }
+        return match if match
+      end
+
+      (rows.find { |r| r[3].to_s.casecmp('id').zero? } || rows.first)&.dig(3)
     end
 
     def execute_to_hashes(sql, binds, projected_columns)
--- a/packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/collection.rb
+++ b/packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/collection.rb
@@ -12,7 +12,6 @@ module ForestAdminDatasourceSnowflake
     def initialize(datasource, table_name)
       super
       @table_name = table_name
-      @primary_keys = []
       @json_columns = []
       fetch_fields
       enable_count
@@ -77,7 +76,6 @@ module ForestAdminDatasourceSnowflake
           enum_values: [],
           validation: nullable ? [] : [{ operator: 'Present' }]
         )
-        @primary_keys << column_name if field.is_primary_key
         add_field(column_name, field)
       end
     end
--- a/packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/collection.rb
+++ b/packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/collection.rb
@@ -55,7 +55,7 @@ module ForestAdminDatasourceSnowflake
       end
 
       native_types = @datasource.fetch_snowflake_native_types(@table_name)
-      pk_name = resolve_primary_key(rows)
+      pk_names = resolve_primary_keys(rows)
 
       rows.each do |row|
         column_name = row[3]
@@ -69,7 +69,7 @@ module ForestAdminDatasourceSnowflake
         field = ForestAdminDatasourceToolkit::Schema::ColumnSchema.new(
           column_type: forest_type,
           filter_operators: Parser::Column.operators_for_column_type(forest_type),
-          is_primary_key: column_name == pk_name,
+          is_primary_key: pk_names.include?(column_name),
           is_read_only: true,
           is_sortable: true,
           default_value: nil,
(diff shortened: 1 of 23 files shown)


matthv:
The PR is missing changes to `.releaserc.js`. The new forest_admin_datasource_snowflake package needs to be wired into 3 sections so it gets published on the next release:                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                            
  - prepareCmd                                                                                                                         
  - successCmd                                                                                                                                                                                                             
  - assets                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                            
  Without this, the gem won't be published to RubyGems even after the PR is merged.

packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/collection.rb
```diff
@@ -0,0 +1,129 @@
...
+        begin
+          stmt.fetch_all || []
+        ensure
+          stmt.drop
+        end
+      end
+
+      native_types = @datasource.fetch_snowflake_native_types(@table_name)
+      pk_name = (rows.find { |r| r[3].to_s.casecmp('id').zero? } || rows.first)&.dig(3)
+
```
matthv, line 50: What is the default order returned by `conn.columns` when no column is named “id”? Couldn't `rows.first` lead to unexpected behavior?

packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/datasource.rb
```diff
@@ -0,0 +1,165 @@
...
+      driver = ::ODBC::Driver.new
+      driver.name  = 'odbc'
+      driver.attrs = attrs
+      conn = ::ODBC::Database.new.drvconnect(driver)
+      apply_session_settings(conn)
+      conn
+    end
+
+    def apply_session_settings(conn)
+      run_session_statement(conn, "ALTER SESSION SET TIMEZONE = 'UTC'")
```
matthv, line 184: I might be mistaken (I've only recently started working with Snowflake), but I've noticed a synchronization issue around schema configuration.
In initialize, the schema can be defined via the Ruby schema: option. However, Snowflake also allows users to pass it directly within the connection string (conn_str) using SCHEMA=....

Currently, these two methods are out of sync:
- visible_tables relies on @schema_override.
- fetch_snowflake_native_types and Parser::Relation.discover both rely on the session's CURRENT_SCHEMA().

 The issue: If a user only sets the Ruby schema: option, the session remains unaware of it. As a result:
  - JSON columns are incorrectly rendered as String (the Snowflake ODBC driver returns VARIANT/OBJECT/ARRAY as SQL_VARCHAR unless the native type lookup corrects them — but that lookup itself queries the wrong schema).
  - Foreign keys are not discovered.
  - No error is raised, which makes this hard to debug.

Proposed fix: Adding a USE SCHEMA in apply_session_settings (when @schema_override is set)

packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/collection.rb
bexchauveto: The order is not guaranteed by the ODBC spec. The Snowflake ODBC driver happens to return columns in ORDINAL_POSITION order (the CREATE TABLE definition order), and INFORMATION_SCHEMA confirms that. So rows.first resolves to whatever the operator declared first in the DDL.

That works for tables where id is first by convention, but it's a real foot-gun: a table like (created_at TIMESTAMP, id NUMBER) would silently designate created_at as the PK. Composite primary keys also collapse to a single column with no warning.

The flow now to get the PK is this [commit](https://github.com/ForestAdmin/agent-ruby/pull/292/changes/ce8478e34526eb8351bde52d864021ee464caf01):

1. Operator override — primary_keys: { 'TABLE_NAME' => 'COLUMN_NAME' } constructor option, case-insensitive table lookup
2.  Snowflake-declared PK — SHOW PRIMARY KEYS IN SCHEMA (lowest key_sequence wins for composite PKs)
3. id column (case-insensitive)
4. First column (last resort)

packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/datasource.rb
```diff
@@ -0,0 +1,202 @@
...
+        stmt = conn.prepare('SHOW PRIMARY KEYS IN SCHEMA')
+        rows = begin
+          stmt.execute
+          stmt.fetch_all || []
+        ensure
+          stmt.drop
+        end
+
+        rows.group_by { |r| r[3].to_s.upcase }
+            .transform_values { |table_rows| table_rows.min_by { |r| r[5].to_i }[4].to_s }
```
matthv: Composite PKs (PRIMARY KEY (a, b)) get truncated to their first column here

packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/collection.rb
```diff
@@ -0,0 +1,138 @@
...
+      end
+    end
+
+    def resolve_primary_key(rows)
+      column_names = rows.map { |r| r[3] }
+      declared_pk = @datasource.primary_key_for(@table_name)
+      if declared_pk
+        match = column_names.find { |name| name.to_s.casecmp(declared_pk.to_s).zero? }
+        return match if match
+      end
```
matthv: If the user mistypes the column name in primary_keys:, we just silently fall back to id/first column. They'd have no idea their override was ignored. IMO this should raise: it's a config typo, not a discovery thing where warn makes sense.

Proposed fix : raise ForestAdminDatasourceSnowflake::Error, "primary_keys override '#{declared_pk}' does not match any column on table '#{@table_name}' (available: #{column_names.join(', ')})"

packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/datasource.rb
```diff
@@ -0,0 +1,226 @@
...
+        begin
+          stmt.execute(*binds)
+          rows = stmt.fetch_all || []
+          rows.group_by { |r| r[0].to_s.upcase }
+        ensure
+          stmt.drop
+        end
+      end
+    rescue ::ODBC::Error
+      nil
```
matthv, line 101: This rescues silently to nil. 
If column introspection fails, collections end up with zero fields and Forest crashes later, IMO making the actual cause hard to trace. 
Could we mirror what discover_relations does on line 205?

packages/forest_admin_datasource_snowflake/lib/forest_admin_datasource_snowflake/datasource.rb
```diff
@@ -0,0 +1,226 @@
...
+          stmt.fetch_all || []
+        ensure
+          stmt.drop
+        end
+
+        rows.group_by { |r| r[3].to_s.upcase }
+            .transform_values { |table_rows| table_rows.sort_by { |r| r[5].to_i }.map { |r| r[4].to_s } }
+      end
+    rescue ::ODBC::Error
+      nil
```
matthv, line 132: same silent rescue, same fix as above (fetch_snowflake_columns).

matthv (approved):

PR #303, opened 2026-05-13
feat(zendesk): add plugins for creating tickets and closing actions

christophebrun-forest:
## Summary

Two opt-in plugins for the `forest_admin_datasource_zendesk` package, registered through the standard customizer plugin mechanism (`collection_customizer.add_plugin(plugin_class, options)`). Neither plugin runs by default — covered by [`datasource_spec.rb#registers no smart actions by default`](packages/forest_admin_datasource_zendesk/spec/forest_admin_datasource_zendesk/datasource_spec.rb#L32-L36).

- **`Plugins::CreateTicketWithNotification`** (Single scope). Host-agnostic — Zendesk creates the requester user on the fly from the form's email, so the plugin can attach to any collection. Form fields: `Requester email` (required, pre-filled via `requester_email_default:` — String or Proc), `Subject`, `Message` (RichText → `html_body`), `Priority`, `Type`, `Send as internal note`. Supports `{{record.<field>}}` token interpolation in `default_subject`/`default_message`, HTML-escaped in Message to block injection. Knobs: `action_name`, `priority_override`/`type_override` (force a value and hide the field), `sender_email` (maps to Zendesk `recipient`), `email_templates` (flips the form into a two-page wizard with a Template dropdown), `ticket_id_field` (best-effort writeback of the new ticket id to the host record — a writeback failure is logged and surfaced in the success message without rolling back the Zendesk ticket).
- **`Plugins::CloseTicket`** (Single + Bulk × `solved`/`closed`). Reads the Zendesk ticket id from a configurable host field (`ticket_id_field:`), filters variants via `statuses:` / `scopes:`. Per-id rescue so a single transition rejection (e.g. `open → closed`) doesn't abort the rest of a bulk run; partial successes and failures are surfaced in the result message. Detects Zendesk's "closed prevents ticket update" wrapper to keep "already closed" idempotent for `status=closed` (success), and surfaces a clean error for `status=solved` (cannot reopen).
- **Shared**: ticket enum values (`STATUS`/`PRIORITY`/`TYPE`) extracted to `TicketEnums`, consumed by both the Ticket schema and the form builder.

## Test plan

- [x] `bundle exec rspec` — 220 examples, 0 failures, 99% coverage
- [x] `bundle exec rubocop` — 0 offenses
- [ ] Manual: attach `CreateTicketWithNotification` to a host collection with `requester_email_default:` (Proc), verify the email pre-fills from the selected record
- [ ] Manual: configure `email_templates:` and confirm the two-page wizard, Template selection, and `{{record.<field>}}` interpolation in the rendered Message
- [ ] Manual: attach `CloseTicket` with `statuses: %w[solved closed]`, run a bulk close on a mix including one ticket in a state Zendesk rejects, verify the others still transition and the failed id is reported
- [ ] Manual: configure `ticket_id_field:` on `CreateTicketWithNotification`, create a ticket, verify the host record's column is updated; then force the writeback to fail and verify the ticket is still created and the warning surfaces in the success message

> [!NOTE]
> ### Add smart actions for Zendesk ticket creation and closure to the Zendesk datasource
> - Adds a `CloseTicket` plugin that registers smart actions (single and bulk scope) to set Zendesk ticket status to `solved` or `closed`, reporting granular success/failure and treating already-closed tickets idempotently.
> - Adds a `CreateTicketWithNotification` plugin that registers a single-scope action to create a Zendesk ticket from a form, optionally notify the requester, and write the new ticket ID back to the host record.
> - The create action form supports record-based field interpolation, an optional two-page template selection wizard, and safe HTML handling for the message body.
> - Extracts shared ticket enum constants (`STATUS`, `PRIORITY`, `TYPE`) into a new `TicketEnums` module consumed by both plugins and the existing schema definition.
> - Adds `forest_admin_datasource_customizer` as a runtime dependency; both plugins inherit from its `Plugin` base class.
>
> 
> #### Changes since #303 opened
>
> - Introduced validation of API response structure in write operations [42789c9]
> - Replaced `ArgumentError` with `ForestException` for validation errors across plugins [42789c9]
> - Removed symbol key support from record field access [42789c9]
> - Refactored 'already closed' ticket detection to use structured error inspection [42789c9]
> - Enabled template token interpolation for requester email defaults [42789c9]
> - Enforced hash response structure from ticket creation operations [42789c9]
> - Updated test expectations to match new error types and key access patterns [42789c9]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8f6540d.</sup>
> 
>
>


packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/plugins/close_ticket.rb
```diff
@@ -0,0 +1,131 @@
...
+              "#{e.class}: #{e.message}"
+            )
+            failed << [id, "#{e.class}: #{e.message}"]
+          end
+        end
+        [succeeded, already_closed, failed]
+      end
+
+      def already_closed?(error)
+        error.message.to_s.include?(ALREADY_CLOSED_PATTERN)
```
matthv: Relies on a brittle string match. We could match via e.cause (the structured RecordInvalid) instead, IMO it's more robust

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/plugins/close_ticket.rb
```diff
@@ -0,0 +1,131 @@
...
+      }.freeze
+
+      SCOPES = { single: ActionScope::SINGLE, bulk: ActionScope::BULK }.freeze
+
+      def run(_datasource_customizer, collection_customizer = nil, options = {})
+        datasource = options[:datasource]
+        ticket_id_field = options[:ticket_id_field]
+        raise ArgumentError, 'CloseTicket plugin requires :datasource' unless datasource
+        raise ArgumentError, 'CloseTicket plugin requires :ticket_id_field' unless ticket_id_field
+        raise ArgumentError, 'CloseTicket plugin requires a collection' unless collection_customizer
```
matthv: Could we unify the config-error raises in #run to ForestException (currently ArgumentError)? That's what Plugins::AddExternalRelation does for the same case, and it lets the caller rescue a single type. The plugins already mix both types depending on the validation step, so no real reason to keep the split.

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/plugins/close_ticket.rb
```diff
@@ -0,0 +1,131 @@
...
+            result_builder.error(message: Messages.error(failed, status))
+          else
+            result_builder.success(message: Messages.success(succeeded, already_closed, failed, status))
+          end
+        end
+      end
+
+      def resolve_ticket_ids(context, ticket_id_field)
+        records = context.get_records([ticket_id_field])
+        records = [records].compact unless records.is_a?(Array)
```
matthv: `the records = [records].compact unless records.is_a?(Array)` guard can never trigger, get_records delegates to `collection.list` which always returns an Array. Any reason to keep it, or can we just drop it?

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/plugins/close_ticket.rb
```diff
@@ -0,0 +1,131 @@
...
+        records.filter_map { |r| r[ticket_id_field] || r[ticket_id_field.to_sym] }
```
matthv: The r[ticket_id_field] || r[ticket_id_field.to_sym] fallback in resolve_ticket_ids, all Forest datasources return string-keyed records. The symbol fallback covers a scenario that doesn't happen in the framework. Could we simplify to r[ticket_id_field.to_s] and drop the "works with symbol keys" test?

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/plugins/create_ticket_with_notification/form_builder.rb
```diff
@@ -0,0 +1,149 @@
...
+          )
+          {}
+        end
+
+        # Message ships as html_body — unescaped `<` or `&` from a record
+        # value would break the outbound email or smuggle markup into it.
+        def interpolate(template, record, escape_html:)
+          template.gsub(TOKEN_RE) do
+            key = ::Regexp.last_match(1)
+            value = record[key] || record[key.to_sym]
```
matthv: Same string/symbol issue as resolve_ticket_ids. Same fix here.

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/plugins/create_ticket_with_notification.rb
```diff
@@ -0,0 +1,107 @@
...
+
+      def executor(datasource, opts)
+        lambda do |context, result_builder|
+          values = context.form_values
+          email  = values['Requester email']
+          next result_builder.error(message: 'Requester email is required.') unless present?(email)
+
+          payload = build_payload(values, email, opts)
+          ticket = datasource.client.create_ticket(payload)
+          ticket_id = ticket.respond_to?(:[]) ? ticket['id'] : nil
```
matthv: IMO `The ticket.respond_to?(:[]) ? ticket['id'] : nil` masks a contract issue with `Client#create_ticket.` If the return shape isn't guaranteed, that should be fixed in writes.rb rather than guarded here. The silent nil makes the success message say "Ticket created" without an id, which is hard to debug. 
Could we either drop the guard and let ticket['id'] raise naturally if the shape is wrong, or check explicitly and raise a typed error with context?

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/plugins/create_ticket_with_notification/form_builder.rb
```diff
@@ -0,0 +1,149 @@
...
+
+        def internal_note_field
+          { type: FieldType::BOOLEAN, label: 'Send as internal note',
+            description: 'When checked, the first comment is private and no email is sent to the requester.',
+            default_value: false }
+        end
+
+        def requester_default(value)
+          return nil if value.nil?
+          return value if value.is_a?(String)
```
matthv: WDYT about running `template_default(value, escape_html: false) when value.is_a?(String)`. Aligns it with the other String defaults' token support.

matthv (commented):
Given how the PR has evolved, I suggest a slightly more appropriate title: `feat(zendesk): add plugins for creating tickets upon registration and closing actions`

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/plugins/close_ticket.rb
christophebrun-forest: yes drop it

packages/forest_admin_datasource_zendesk/lib/forest_admin_datasource_zendesk/plugins/create_ticket_with_notification/form_builder.rb
christophebrun-forest: right, done

matthv (approved):
LGTM

PR #307, opened 2026-05-20
fix(rpc): action with file

arnaud-moncel:
### General

### Security

> [!NOTE]
> ### Fix RPC action execute to return file payloads with correct headers and binary body
> - Adds `call_rpc_raw` to `RpcClient` that returns the raw `Faraday::Response` so callers can inspect headers and status, while `call_rpc` continues to return parsed body.
> - `Collection#execute` now calls `call_rpc_raw` and checks the `x-forest-action-type` header; when it equals `'File'`, it builds a structured file result via `build_file_result` with mime type, decoded filename, raw stream, and optional response headers.
> - `ActionExecute#handle_request` detects file results and returns a raw binary HTTP response with `Content-Type`, `Content-Disposition`, and `X-Forest-Action-*` headers instead of JSON-encoding the result.
> - `BaseRoute` (both Sinatra and Rails paths) gains support for a `:raw` flag in the response hash, bypassing JSON serialization so binary content is returned as-is.
> - Behavioral Change: action endpoints that previously returned file results as JSON now return raw binary responses with file-specific HTTP headers.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 822b9bb.</sup>
> 
>

### DIFF
--- a/packages/forest_admin_datasource_rpc/lib/forest_admin_datasource_rpc/Utils/rpc_client.rb
+++ b/packages/forest_admin_datasource_rpc/lib/forest_admin_datasource_rpc/Utils/rpc_client.rb
@@ -31,15 +31,22 @@ module ForestAdminDatasourceRpc
       end
 
       # rubocop:disable Metrics/ParameterLists
-      def call_rpc(endpoint, caller: nil, method: :get, payload: nil, symbolize_keys: false,
-                   if_none_match: nil, with_response: false)
+      def call_rpc(endpoint, caller: nil, method: :get, payload: nil, symbolize_keys: false, if_none_match: nil)
+        resp = call_rpc_raw(endpoint, caller: caller, method: method, payload: payload,
+                                      symbolize_keys: symbolize_keys, if_none_match: if_none_match)
+        return resp if resp == NotModified
+
+        resp.body
+      end
+
+      def call_rpc_raw(endpoint, caller: nil, method: :get, payload: nil, symbolize_keys: false, if_none_match: nil)
         response = make_request(endpoint, caller: caller, method: method, payload: payload,
                                           symbolize_keys: symbolize_keys, if_none_match: if_none_match)
         return NotModified if response.status == HTTP_NOT_MODIFIED
 
         raise_appropriate_error(response) unless response.success?
 
-        with_response ? response : response.body
+        response
       end
       # rubocop:enable Metrics/ParameterLists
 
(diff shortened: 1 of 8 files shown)


matthv:
Hey @arnaud-moncel, I tested this PR with a real PDF (via a `download pdf` action that returns a binary PDF and an `upload pdf` action that echoes back an uploaded PDF). It crashes with:

```
JSON::GeneratorError ("\xE2" from ASCII-8BIT to UTF-8)
  caused by: Encoding::UndefinedConversionError

base_route.rb:104 → BaseRoute#serialize_response
base_route.rb:71  → BaseRoute#build_rails_response
```

### Why

`BaseRoute#serialize_response` runs `result.to_json` on anything that isn't already a JSON string. On binary content (PDF header `%\xE2\xE3\xCF\xD3`, or any byte > 0x7F in an image/archive), `ActiveSupport::JSON.encode` tries to produce valid UTF-8 → `Encoding::UndefinedConversionError` → 500.

The bug predates this PR. On `main`, the whole File hash gets JSON-encoded (crashes on the value of `:stream`). With this PR, `result[:content]` gets JSON-encoded directly (crashes on the same bytes). Same root cause, just one frame deeper in the stack — `serialize_response` still ends up called on binary in both cases.

I confirmed both setups: reproduced on `main` HEAD with a direct curl, reproduced on this branch with my PDF generator and with a real user PDF via `upload pdf`.

### Fix (commit 1b6e6b546)

I added a `raw: true` flag to the response hash you build in `ActionExecute#build_file_response`. In `BaseRoute` (both `register_sinatra` and `build_rails_response` paths), when `result[:raw]` is true we pass `result[:content].to_s` to Rack as-is instead of calling `serialize_response`. Bytes arrive intact, ASCII-8BIT encoding is preserved.

I also added the tests that were missing on the agent side: `action_execute_spec.rb` didn't cover the File path at all (the datasource specs passed with a `body: 'binary-payload'` mock that doesn't exercise the Rack serialization step — that's what hid the original bug). We now have:
- `BaseRoute#build_rails_response` with a `raw: true` binary → intact bytes, encoding preserved
- `ActionExecute#handle_request` returns `{status:, raw: true, content: <binary>, headers: {...}}` for a File result
- coverage for both with/without `response_headers`

### Points not addressed in this commit (worth deciding before merge)

- `Content-Disposition: attachment; filename="..."` contains the URL-encoded name (spaces → `+`). If a browser reads it directly, it downloads `report+final.pdf`. Since the RPC client reads `X-Forest-Action-File-Name` instead, this header is effectively dead code in the RPC flow. Either fix to RFC 5987 (`filename*=UTF-8''...`) or drop it.
- `X-Forest-Action-Response-Headers: <JSON>` works for simple cases (`Set-Cookie: token=xyz`), but a real cookie with `Path=/; HttpOnly; Secure` or CR/LF in the value will exceed HTTP header limits. Worth deciding whether we keep this mechanism or move those metadata into the body.
- The Macroscope summary mentions schema changes (`actions` map, `BaseAction.static_form`) that aren't in the diff — is that in another PR, or is the auto-summary just off?

Let me know if you want me to handle the first two in this PR or split them off into a follow-up.

matthv (approved):

PR #310, opened 2026-05-22
fix(rpc agent): allow connect as node rpc

arnaud-moncel:
### General

### Security

> [!NOTE]
> ### Fix RPC agent Rails routes to include URL path parameters in request params
> - Adds `extract_request_params` helper in [`BaseRoute`](https://github.com/ForestAdmin/agent-ruby/pull/310/files#diff-0b1b029dc4c0c7c8eadc71e2943a8339c4770b4d0ed1a1a25527f297b3899f1a) that merges `path_parameters`, `query_parameters`, and `request_parameters` so that URL segment params like `:collection_name` are available to route handlers.
> - Route handlers in `list`, `create`, `aggregate`, and `action_form` previously received only query/body params, causing `collection_name` to be missing when provided as a path segment.
> - Changes the missing `collection_name` fallback return value from `{}` to `[]` across all affected handlers.
>
> 
> #### Changes since #310 opened
>
> - Removed early-return guards from route handlers to raise errors for missing or unknown collections [8e3ff47]
> - Refactored parameter extraction to use indifferent access without deep key symbolization [8e3ff47]
> - Updated route handler specs to expect errors instead of empty results for unknown collections [8e3ff47]
> - Added test coverage for `ForestAdminRpcAgent::Routes::Aggregate#handle_request` to verify that when the aggregation payload omits the `:groups` key, `Aggregation.new` is invoked with `groups` defaulting to an empty array [e0ffe0d]
> - Added test coverage for `ForestAdminRpcAgent::Routes::BaseRoute#extract_request_params` to verify parameter merging behavior from path, query, and body sources with indifferent access and proper filtering of Rails-internal keys [e0ffe0d]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized cc8ced2.</sup>
> 
>
>

### DIFF
--- a/packages/forest_admin_rpc_agent/lib/forest_admin_rpc_agent/routes/aggregate.rb
+++ b/packages/forest_admin_rpc_agent/lib/forest_admin_rpc_agent/routes/aggregate.rb
@@ -12,8 +12,6 @@ module ForestAdminRpcAgent
       end
 
       def handle_request(args)
-        return [] unless args[:params]['collection_name']
-
         datasource = ForestAdminRpcAgent::Facades::Container.datasource
         collection = get_collection_safe(datasource, args[:params]['collection_name'])
 
--- a/packages/forest_admin_rpc_agent/lib/forest_admin_rpc_agent/routes/aggregate.rb
+++ b/packages/forest_admin_rpc_agent/lib/forest_admin_rpc_agent/routes/aggregate.rb
@@ -12,7 +12,7 @@ module ForestAdminRpcAgent
       end
 
       def handle_request(args)
-        return {} unless args[:params]['collection_name']
+        return [] unless args[:params]['collection_name']
 
         datasource = ForestAdminRpcAgent::Facades::Container.datasource
         collection = get_collection_safe(datasource, args[:params]['collection_name'])
@@ -20,7 +20,7 @@ module ForestAdminRpcAgent
         aggregation = Aggregation.new(
           operation: args[:params]['aggregation']['operation'],
           field: args[:params]['aggregation']['field'],
-          groups: args[:params]['aggregation']['groups']
+          groups: args[:params]['aggregation']['groups'] || []
         )
         filter = FilterFactory.from_plain_object(args[:params]['filter'])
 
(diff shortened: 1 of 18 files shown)


packages/forest_admin_rpc_agent/lib/forest_admin_rpc_agent/routes/create.rb
```diff
@@ -8,7 +8,7 @@ def initialize
       end
 
       def handle_request(args)
-        return {} unless args[:params]['collection_name']
+        return [] unless args[:params]['collection_name']
```
matthv: Inconsistency: the return value was changed from {} to [] in list, create, aggregate, action_form, but the same line still returns {} in update, delete, action_execute, chart, and native_query. Any reason for treating these differently?

matthv: With the new extract_request_params merging path_parameters, collection_name is now always populated when the request comes through Rails routing.
This early return is now effectively dead code on the Rails path WDYT?
Could we just remove the guard? If collection_name can no longer be missing in practice, the cleanest move is to drop the line entirely

packages/forest_admin_rpc_agent/lib/forest_admin_rpc_agent/routes/aggregate.rb
```diff
@@ -12,15 +12,15 @@ def initialize
...
+        return [] unless args[:params]['collection_name']
 
         datasource = ForestAdminRpcAgent::Facades::Container.datasource
         collection = get_collection_safe(datasource, args[:params]['collection_name'])
 
         aggregation = Aggregation.new(
           operation: args[:params]['aggregation']['operation'],
           field: args[:params]['aggregation']['field'],
-          groups: args[:params]['aggregation']['groups']
+          groups: args[:params]['aggregation']['groups'] || []
```
matthv, line 21: Add a quick test in aggregate_spec.rb covering the case aggregation is passed without groups, to lock the behavior

matthv (commented):
Nice catch on the missing path_parameters merge.

I suggest to add a regression test for extract_request_params in base_route_spec.rb.
The specs were updated from eq({}) to eq([]) but none actually test the scenario being fixed (collection_name resolved from path_parameters only)

matthv (approved):
LGTM 🚀

PR #321, opened 2026-06-25
feat(agent): forward any executor route generically [PRD-567]

matthv:
## What

Ports the generic executor passthrough to the Ruby agent, matching the agent-nodejs reference ([agent-nodejs#1707](https://github.com/ForestAdmin/agent-nodejs/pull/1707), released 1.81.0).

- **Single catch-all route** `_internal/executor/*path` → executor `/:path` **verbatim** (all verbs). Replaces the `/runs`-confined proxy. Callers address runs as `runs/<id>`, so any executor route (mcp-oauth-credentials, future ones) is reachable with no agent change.

## Security

- **SSRF guard**: string guard (empty / leading-`/` / `..` / `%2e`/`%2E` / backslash / NUL → `NotFoundError`) plus an authoritative origin check (`URI.parse`, reject unless scheme+host+port match the executor). A decoded control char like `\t//host` makes `URI.parse` raise → rejected.
- Faraday's client has no FollowRedirects middleware, so redirects aren't followed (matches the Node agent's raw http client).

## Tests

**24/24** + rubocop clean. Run GET (caller includes `runs/`), POST trigger, **non-runs route verbatim** (asserts no `/runs` prefix injected), header req/resp filtering, status/body passthrough, SSRF vectors (`..`, `%2e`, NUL, control-char `\t//evil.com`) → `NotFoundError`, unreachable/timeout/SSL → 503.

fixes PRD-567

> [!NOTE]
> ### Forward executor routes generically under `/_internal/executor` in `WorkflowExecutorProxy`
> - Changes the agent mount path from `/_internal/workflow-executions` to `/_internal/executor` and removes the hardcoded `/runs` prefix injection, so paths are forwarded verbatim to the configured executor URL.
> - Adds `reject_off_origin!` to block SSRF by comparing the computed target URL's scheme, host, and port against the configured executor origin, raising `NotFoundError` on mismatch.
> - Callers must now include the full subpath (e.g. `runs/...`) in the request — the proxy no longer prepends it.
> - Risk: any existing integration that relied on the `/_internal/workflow-executions` mount point or the automatic `/runs` prefix will break without a corresponding caller update.
>
> #### 🖇️ Linked Issues
> Implements [PRD-567](ticket:linear/PRD-567).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized eafae3b.</sup>
> 
>

### DIFF
--- a/packages/forest_admin_agent/lib/forest_admin_agent/routes/workflow/workflow_executor_proxy.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/routes/workflow/workflow_executor_proxy.rb
@@ -1,13 +1,13 @@
 require 'faraday'
+require 'uri'
 
 module ForestAdminAgent
   module Routes
     module Workflow
-      # Generic proxy: forwards any sub-path/verb under AGENT_PREFIX to EXECUTOR_PREFIX, so a
-      # new executor route needs no change here (PRD-567). Mounted only when `workflow_executor_url` is set.
+      # Generic proxy: forwards any sub-path/verb under AGENT_PREFIX to the executor verbatim, so a
+      # new executor route needs no change here. Mounted only when `workflow_executor_url` is set.
       class WorkflowExecutorProxy < AbstractAuthenticatedRoute
-        AGENT_PREFIX = '/_internal/workflow-executions'.freeze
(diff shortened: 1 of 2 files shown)


christophebrun-forest (approved):
LGTM

PR #333, opened 2026-07-21
fix(forest_admin_agent): build PolymorphicManyToOne linkage id from foreign key (#332)

### LINKED ISSUE #332: PolymorphicManyToOne serializes an empty id in list view → frontend crashes with "Expected id to be a string or number, received null"
## Environment
- `forest_admin_agent` **1.35.3**
- `forest_admin_datasource_active_record` **1.35.3**
- `forest_admin_rails` **1.35.3**
- Ruby 4.0.2 / Rails 8.1, PostgreSQL
- `support_polymorphic_relations: true` enabled on the datasource

## Summary
When a collection with a **polymorphic `belongs_to`** contains at least one record, opening its **list view** fails on the frontend with:

```
UnknownClientRecordAgentError: Expected id to be a string or number, received null
  at O.fetchRecords (feature.ts:440)
{requestType: 'list'}
```

The list becomes completely unusable ("An error occurred. Please try again…"). The backend returns **200 OK** — the payload itself is malformed.

## Reproduction
Models:
```ruby
class Address < ApplicationRecord
  belongs_to :addressable, polymorphic: true
end

class Contact < ApplicationRecord
  has_one :address, as: :addressable
end
```
Create one `Address` linked to a `Contact`, then open the **Address** list in the UI.

## Root cause
For a `PolymorphicManyToOne`, `QueryStringParser#build_projection_fields` builds a wildcard sub-projection:
```ruby
# forest_admin_agent/utils/query_string_parser.rb:72
when 'PolymorphicManyToOne'
  "#{field_name}:*"
```
But a polymorphic `belongs_to` can't be SQL-joined, so the AR datasource **skips** it while building the SELECT:
```ruby
# forest_admin_datasource_active_record/utils/query.rb:208
next if relation_schema.type == 'PolymorphicManyToOne'
```
The record therefore comes back with a **phantom** related object that has no primary key:
```ruby
{ "addressable_id" => 1, "addressable_type" => "Contact", "addressable" => { "*" => nil } }
```
The serializer then reads the pk from that phantom record, producing an **empty id** in two places:

1. Relationship linkage:
```json
"addressable": { "data": { "type": "Contact", "id": "" } }
```
2. Compound `included` document (note the empty self-link):
```json
"included": [{ "type": "Contact", "id": "", "links": { "self": "/forest/Contact/" } }]
```

EmberData rejects the empty id → the whole list fails to render.

## Expected behavior
The linkage `id` (and the `included` resource id, if any) should be built from the **foreign key already present on the owner record** (`addressable_id` + `addressable_type`), which does not require loading the related record.

## Notes
- Only reproduces when the collection has ≥ 1 record (empty polymorphic collections render fine, which masks the bug).

## Workaround (in our app, not a gem fix)

Since the related record's primary key can't come from a SQL join for a polymorphic
`belongs_to`, we rebuild it from the foreign key already present on the owner record.
No extra query is issued (the FK is already loaded in memory):

```ruby
ForestAdminAgent::Serializer::ForestSerializer.prepend(Module.new do
  def has_one_relationship(attribute_name, attr_data)
    object = super
    return object unless object.is_a?(Hash)

    collection = ForestAdminAgent::Facades::Container.datasource.get_collection(
      @options[:class_name].gsub("::", "__")
    )
    field = collection.schema[:fields][attribute_name.to_s]
    return object unless field&.type == "PolymorphicManyToOne"

    foreign_key = @object[field.foreign_key]
    foreign_type = @object[field.foreign_key_type_field]
    return nil if foreign_key.nil? || foreign_type.nil?

    primary_key = (field.foreign_key_targets || {})[foreign_type] || "id"
    object.merge(primary_key => foreign_key)
  end
end)
```

Verified: fixes both the relationship linkage and the `included` document, no N+1
(0 extra SQL queries measured serializing 50 polymorphic rows), and unlinked
records (`foreign_key` nil) correctly omit the `data` key instead of crashing.

The proper fix probably belongs in `ForestSerializer#has_one_relationship` or in the
datasource's projection handling for `PolymorphicManyToOne` (`query.rb:208`), so the
sub-projection isn't silently skipped.

bexchauveto:
Fixes #332

## Problem
Opening the list view of a collection with a polymorphic `belongs_to` crashes the frontend:

```
UnknownClientRecordAgentError: Expected id to be a string or number, received null
```

The backend returns 200 OK but the payload is malformed: the relationship linkage `id` is an empty string.

## Root cause
A `PolymorphicManyToOne` can't be SQL-joined, so `query.rb:208` skips it while building the SELECT. The record therefore comes back with a **phantom** related object that has no primary key (`{ "*" => nil }`). `ForestSerializer` then reads the PK off that phantom object, producing an empty id in the JSON:API linkage.

The existing serializer specs didn't catch this because they hand-fed a fully-populated related hash, which the real datasource never produces.

## Fix
Override `has_one_relationship` in `ForestSerializer`. For a `PolymorphicManyToOne` it rebuilds the linkage id from the foreign key (`*_id` + `*_type`) already loaded on the owner record — no extra query — and returns `nil` (omitting the `data` key) when the relation is unlinked. Both the linkage and the compound `included` document go through this method, so a single override covers both.

This is the workaround reported in #332, moved into the gem.

## Tests
Added two specs reproducing the real datasource output (phantom related object) — both fail without the fix:
- builds the linkage id from the foreign key when the related record is a phantom
- omits the `data` key for an unlinked polymorphic relation

> [!NOTE]
> ### Fix `PolymorphicManyToOne` linkage id serialization in `ForestSerializer`
> When a polymorphic many-to-one related record is a phantom (no primary key), the serializer now rebuilds the linkage `id` from the owner's foreign key fields instead of returning an empty or broken relationship.
>
> - Overrides `has_one_relationship` in [`ForestSerializer`](https://github.com/ForestAdmin/agent-ruby/pull/333/files#diff-b79f280368f0a5ff9ed78fbf1be38a2a2b042e30672ddc1244b31713fae3ffd6) to post-process `PolymorphicManyToOne` relationships, computing `id` from the foreign key value and resolving the type using `foreign_key_targets`.
> - Namespaced type values (e.g. `Admin::User`) are formatted to `Admin__User` to match collection naming conventions, and custom primary keys are respected.
> - Relationships with a nil foreign key or nil type omit the `data` member entirely.
> - Risk: any consumer relying on the previous (broken) phantom linkage behavior will see different output.
>
> 
> #### Changes since #333 opened
>
> - Added test verifying PolymorphicManyToOne relationship serialization [618ae03]
> - Clarified comments for PolymorphicManyToOne relationship handling [618ae03]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 870bad8.</sup>
> 
>
>

### DIFF
--- a/packages/forest_admin_agent/lib/forest_admin_agent/serializer/forest_serializer.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/serializer/forest_serializer.rb
@@ -112,7 +112,9 @@ module ForestAdminAgent
         foreign_type = @object[field.foreign_key_type_field]
         return nil if foreign_key.nil? || foreign_type.nil?
 
-        primary_key = (field.foreign_key_targets || {})[foreign_type] || 'id'
+        # foreign_key_targets is keyed by formatted collection names (Admin::User => Admin__User),
+        # but foreign_type is the raw AR type column value, so format it before the lookup.
+        primary_key = (field.foreign_key_targets || {})[foreign_type.gsub('::', '__')] || 'id'
         (object.is_a?(Hash) ? object : {}).merge(primary_key => foreign_key)
       end
 
--- a/packages/forest_admin_agent/lib/forest_admin_agent/serializer/forest_serializer.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/serializer/forest_serializer.rb
@@ -99,6 +99,10 @@ module ForestAdminAgent
       # the frontend). No extra query is issued.
       def has_one_relationship(attribute_name, attr_data)
         object = super
+        # nil means the relation wasn't projected/loaded at all (e.g. update/store responses) — keep
+        # the previous behavior. Only the phantom object returned by list projections needs rebuilding.
+        return object if object.nil?
+
         field = ForestAdminAgent::Facades::Container.datasource
                                                     .get_collection(@options[:class_name].gsub('::', '__'))
                                                     .schema[:fields][attribute_name.to_s]
--- a/packages/forest_admin_agent/lib/forest_admin_agent/serializer/forest_serializer.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/serializer/forest_serializer.rb
@@ -93,6 +93,25 @@ module ForestAdminAgent
         @to_one_associations[name] = format_field(name, options)
       end
 
+      # For a PolymorphicManyToOne the related record can't be SQL-joined, so the datasource returns
+      # a phantom object without a primary key. Rebuild the id from the foreign key already loaded on
+      # the owner record instead of reading it off that phantom (which yields an empty id and crashes
+      # the frontend). No extra query is issued.
+      def has_one_relationship(attribute_name, attr_data)
+        object = super
+        field = ForestAdminAgent::Facades::Container.datasource
+                                                    .get_collection(@options[:class_name].gsub('::', '__'))
+                                                    .schema[:fields][attribute_name.to_s]
+        return object unless field&.type == 'PolymorphicManyToOne'
+
+        foreign_key = @object[field.foreign_key]
+        foreign_type = @object[field.foreign_key_type_field]
+        return nil if foreign_key.nil? || foreign_type.nil?
+
+        primary_key = (field.foreign_key_targets || {})[foreign_type] || 'id'
+        (object.is_a?(Hash) ? object : {}).merge(primary_key => foreign_key)
+      end
+
       def has_one_relationships
         return {} if @to_one_associations.nil?
         data = {}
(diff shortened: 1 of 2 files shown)


matthv (commented):
Nit: comments a bit verbose. The foreign_key_targets keying one justifies the gsub → keep. But # nil means… mostly restates return object, and no extra query is issued. reads as pre-empting a reviewer. Could trim both, wdyt?


The specs only assert the linkage, not the included doc (#332's 2nd symptom, the empty self-link). Could be good to add one with include: 'documentable' asserting included has id => '10'?

bexchauveto:
Both done in 618ae032:

- Trimmed the comments — dropped the `nil means…` block (now a one-line inline note) and `No extra query is issued`; kept the `foreign_key_targets` note since the gsub isn't obvious.
- Added a spec that serializes with `include: 'documentable'` and asserts the `included` resource has `id => '10'` and `self => /forest/Car/10` (the empty self-link symptom). Verified it fails without the fix (`got nil`).

Full suite green (724 examples).

matthv (approved):
ok for me 👍

PR #342, opened 2026-08-04
fix(agent): store the model name in the polymorphic type column on create

PMerlet:
## Problem

When a record is created through the JSON:API `relationships` payload and the relation is a `PolymorphicManyToOne`, the agent stored the **Forest collection name** in the polymorphic type column instead of the Rails class name. For a namespaced model, the column ended up holding `Banking__Account` where `Banking::Account` was expected.

Only namespaced models are affected — `gsub('__', '::')` is a no-op on the others, which is why this went unnoticed.

Consequences for a namespaced target:

1. Rails can no longer resolve the association: `commentable_type = 'Banking__Account'` is not a valid constant.
2. The record never shows up in the parent's related-data tab, since `PolymorphicOneToMany` filters on the model name.
3. It looks fine in the detail view: the serializer resolves the target by applying the reverse substitution, which is idempotent on an already-formatted value. So the row is corrupted while the UI seems healthy.

## Fix

`format_attributes` now converts the name back before storing it, like every other write path already did (`update_related`, `associate_related`, `delete`, `store`). The local variable was renamed `collection_name`, since it holds a collection name — naming it `model_name` is what hid the missing conversion in the first place.

The `|| json_api_type` fallback for unmatched types is **deliberately unchanged**: it is load-bearing for legitimate case variants (a frontend sending `Company` for a collection named `company`), so removing it would break existing flows. Hardening it (case-insensitive matching, then 400 on unknown types) is tracked in [PRD-874](https://linear.app/forestadmin/issue/PRD-874/ruby-agent-createupdate-persists-client-supplied-relationship-values).

## Tests

A spec on a namespaced collection was added next to the existing polymorphic one. It fails without the fix with `"memberable_type" => "Banking__Account"`, and passes with it. The 726 specs of `forest_admin_agent` pass, rubocop is clean.

fixes PRD-870

> [!NOTE]
> ### Fix polymorphic type column to store namespaced model name on create
> When creating a record with a `PolymorphicManyToOne` relationship, the type column now stores the namespaced model name (e.g. `Foo::Bar`) by replacing `__` with `::` in the matched collection name. The fallback to the raw `json_api_type` when no foreign collection matches is unchanged (deliberately retained — see PRD-874). The safe navigation added in 6dab1da means a `{ data: nil }` payload now clears both columns instead of raising.
>
> 
> #### Changes since #342 opened
>
> - Updated test spec for polymorphic relationship type storage in `ForestAdminAgent::Routes::Resources` Store [61c148f]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6dab1da.</sup>
> 
>
>

### DIFF
--- a/packages/forest_admin_agent/lib/forest_admin_agent/routes/abstract_authenticated_route.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/routes/abstract_authenticated_route.rb
@@ -24,13 +24,13 @@ module ForestAdminAgent
             record[schema.foreign_key] = value.dig('data', 'id')
             json_api_type = value.dig('data', 'type')
             # Find matching collection from foreign_collections (handles both singular and plural forms)
-            model_name = schema.foreign_collections.find do |coll_name|
+            collection_name = schema.foreign_collections.find do |coll_name|
               coll = collection.datasource.get_collection(coll_name)
               coll.name == json_api_type || coll.name.pluralize == json_api_type
             rescue ForestAdminDatasourceToolkit::Exceptions::ForestException
               false
             end || json_api_type
-            record[schema.foreign_key_type_field] = model_name
+            record[schema.foreign_key_type_field] = collection_name.gsub('__', '::')
           end
         end
 
--- a/packages/forest_admin_agent/lib/forest_admin_agent/routes/abstract_authenticated_route.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/routes/abstract_authenticated_route.rb
@@ -30,7 +30,7 @@ module ForestAdminAgent
             rescue ForestAdminDatasourceToolkit::Exceptions::ForestException
               false
             end || json_api_type
-            record[schema.foreign_key_type_field] = collection_name.gsub('__', '::')
+            record[schema.foreign_key_type_field] = collection_name&.gsub('__', '::')
           end
         end
 
--- a/packages/forest_admin_agent/spec/lib/forest_admin_agent/routes/resources/store_spec.rb
+++ b/packages/forest_admin_agent/spec/lib/forest_admin_agent/routes/resources/store_spec.rb
@@ -348,7 +348,7 @@ module ForestAdminAgent
 
             it 'stores the namespaced model name, not the formatted collection name' do
               collection_account = build_collection(
-                name: 'Banking__Account',
+                name: 'Banking__Sepa__Transfer',
                 schema: {
                   fields: {
                     'id' => ColumnSchema.new(
@@ -373,10 +373,10 @@ module ForestAdminAgent
                     'memberable_id' => ColumnSchema.new(column_type: 'Number'),
                     'memberable_type' => ColumnSchema.new(column_type: 'String'),
                     'memberable' => Relations::PolymorphicManyToOneSchema.new(
-                      foreign_collections: ['Banking__Account'],
+                      foreign_collections: ['Banking__Sepa__Transfer'],
                       foreign_key: 'memberable_id',
                       foreign_key_type_field: 'memberable_type',
-                      foreign_key_targets: { 'Banking__Account' => 'id' }
+                      foreign_key_targets: { 'Banking__Sepa__Transfer' => 'id' }
                     )
                   }
                 }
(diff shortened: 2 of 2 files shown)


packages/forest_admin_agent/lib/forest_admin_agent/routes/abstract_authenticated_route.rb
```diff
@@ -24,13 +24,13 @@ def format_attributes(args, collection)
...
             # Find matching collection from foreign_collections (handles both singular and plural forms)
-            model_name = schema.foreign_collections.find do |coll_name|
+            collection_name = schema.foreign_collections.find do |coll_name|
               coll = collection.datasource.get_collection(coll_name)
               coll.name == json_api_type || coll.name.pluralize == json_api_type
             rescue ForestAdminDatasourceToolkit::Exceptions::ForestException
               false
             end || json_api_type
-            record[schema.foreign_key_type_field] = model_name
+            record[schema.foreign_key_type_field] = collection_name.gsub('__', '::')
```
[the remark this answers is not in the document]
PMerlet: Good catch — confirmed and fixed in 6dab1da6. Reproduced first with a spec (`NoMethodError: undefined method 'gsub' for nil`), then guarded with `&.`, so both columns are set back to nil when the relationship carries no data. The spec is part of the commit.

packages/forest_admin_agent/spec/lib/forest_admin_agent/routes/resources/store_spec.rb
```diff
@@ -345,6 +345,117 @@ def respond_to?(arg)
               end
               expect(result[:name]).to eq('member')
             end
+
+            it 'stores the namespaced model name, not the formatted collection name' do
+              collection_account = build_collection(
+                name: 'Banking__Account',
```
hercemer42: Claude Opus 5 (claude-opus-5[1m]): **Should fix** — this fixture's only `__` is a single occurrence, so the spec cannot tell `gsub` from `sub`. Verified: changing L33 to `collection_name&.sub('__', '::')` leaves this file at 8 examples, 0 failures.

A later "simplification" to `sub` would then store `Admin::Reports__Export` for `Admin::Reports::Export`, Rails would fail to resolve the constant, and `forest_serializer.rb:113` would quietly fall back to `'id'` for the primary key — the exact silent corruption PRD-870 is about, green suite and all. One example with a two-level namespaced collection closes the gap.

packages/forest_admin_agent/lib/forest_admin_agent/routes/abstract_authenticated_route.rb
```diff
@@ -24,13 +24,13 @@ def format_attributes(args, collection)
             record[schema.foreign_key] = value.dig('data', 'id')
             json_api_type = value.dig('data', 'type')
             # Find matching collection from foreign_collections (handles both singular and plural forms)
-            model_name = schema.foreign_collections.find do |coll_name|
+            collection_name = schema.foreign_collections.find do |coll_name|
               coll = collection.datasource.get_collection(coll_name)
               coll.name == json_api_type || coll.name.pluralize == json_api_type
             rescue ForestAdminDatasourceToolkit::Exceptions::ForestException
               false
             end || json_api_type
```
hercemer42, line 32: Claude Opus 5 (claude-opus-5[1m]): **Preferential** — PRE-EXISTING, and the ticket's own "Suggested fix" section asked for a look at it: when nothing in `foreign_collections` matches, this fallback writes the raw client-supplied `data.type` into the discriminator column — and now with `__`→`::` applied to it as well.

`find` already handles the singular and plural forms of legitimate targets, so this branch only fires for a type outside the allowed set — arguably the case that should 400 rather than persist. Neither new spec reaches it.

Not a regression, and reasonable to defer — but worth an explicit decision rather than leaving the ticket's second suggestion silently unaddressed.

packages/forest_admin_agent/spec/lib/forest_admin_agent/routes/resources/store_spec.rb
PMerlet: Fixed in 61c148f4 — the fixture now uses a two-level namespace (`Banking__Sepa__Transfer`). Verified the gap is closed by mutation: with `sub` instead of `gsub` the spec now fails (`Banking::Sepa__Transfer`), with `gsub` it passes.

packages/forest_admin_agent/lib/forest_admin_agent/routes/abstract_authenticated_route.rb
PMerlet: Explicit decision: deferring this one. The fallback is not only the unknown-type branch — the `find` comparison is case-sensitive, so when the frontend sends `Company` for a collection named `company` (the case covered by the first polymorphic spec), it's this fallback that makes the create work. Turning it into a 400 would break that legitimate flow; we'd first need case-insensitive matching in the `find`. Pre-existing behavior, out of scope for PRD-870 — tracked in [PRD-874](https://linear.app/forestadmin/issue/PRD-874/ruby-agent-polymorphic-create-fallback-writes-the-raw-client-supplied).

hercemer42:
Claude Opus 5 (claude-opus-5[1m]): **Should fix** — applies to the PR description, not to any line of the diff.

The Macroscope summary in the description states a behavioural change that did not happen:

> the fallback to `json_api_type` is removed — unmatched polymorphic types now write `nil` to the type column instead of the raw incoming value.

`|| json_api_type` is still there — line 32 of `abstract_authenticated_route.rb` at the current head (`61c148f`) — and neither commit touched it: `912e75d` renamed the local and added `.gsub`, `6dab1da` changed `.gsub` to `&.gsub`. Unmatched types still write the raw incoming value, exactly as before.

Worth correcting because the description now contradicts the decision recorded on this PR. [PRD-874](https://linear.app/forestadmin/issue/PRD-874/ruby-agent-createupdate-persists-client-supplied-relationship-values) documents that fallback as **deliberately retained and load-bearing** — `find` is case-sensitive, so a frontend sending `Company` for a collection named `company` depends on it, and removing it as-is breaks that flow. (Confirmed locally: deleting `|| json_api_type` fails the existing spec at `store_spec.rb:291` with `expected "Company", got nil`.)

So the description says the fallback is gone while the ticket says it is kept on purpose. One is the durable record a future reader will trust — editing the description resolves it. No code change involved.

PMerlet:
Good catch — the description is fixed. The Macroscope summary claimed the `|| json_api_type` fallback was removed; it never was, and the Fix section now states explicitly that it is deliberately retained (load-bearing for case variants like `Company` vs `company`), with the hardening tracked in [PRD-874](https://linear.app/forestadmin/issue/PRD-874/ruby-agent-createupdate-persists-client-supplied-relationship-values). The inaccurate 'Behavioral Change' bullet was replaced; the only actual behavioral change besides the `gsub` is that a `{ data: nil }` payload now clears both columns instead of raising (6dab1da).

hercemer42 (approved):

PR #349, opened 2026-08-10
fix(security): enforce authentication and IP whitelist on private routes [PRD-896]

PMerlet:
## Summary

Fixes [PRD-896](https://linear.app/forestadmin/issue/PRD-896/agent-ruby-internalcapabilities-skips-both-authentication-and-the-ip): `POST /forest/_internal/capabilities` answered with **no token verification and no IP whitelist check** — reachable by any anonymous caller. The route inherited from `AbstractRoute` and never called `build(args)`, which is the only place both controls run (there is no auth middleware in agent-ruby).

An adversarial review of the fix extended the scope (documented in the ticket):

- **`Capabilities::Collections`** — inherits `AbstractAuthenticatedRoute` and calls `build(args)`, like every other private route. Note: the ticket's suggested one-liner (base class change alone) is not sufficient — without the `build(args)` call the inheritance protects nothing.
- **`Charts::ApiChartDatasource`** (both handlers) and **`Security::ScopeInvalidation`** — same class of bug: they verified the JWT via `parse_caller` but skipped the IP whitelist. Now aligned on the `build(args)` pattern.
- **`CallerParser`** — a forged, malformed, or expired JWT raised an unrescued `JWT::DecodeError` → 500 "Unexpected error". The front only re-authenticates on 401 (which is what agent-nodejs returns via koa-jwt), so an expired agent token silently degraded capabilities for the whole session. Now rescued → `UnauthorizedError` → 401.
- **CI** — `forest_admin_rails` was linted but never tested (missing from the test matrix, broken `Gemfile-test`, 5 stale `exception_handler` specs asserting a no-log-in-production behaviour removed in #281). Repaired and added to the test matrix + coverage upload, so the controller dispatch and HTTP error mapping are covered again.

## Front-end compatibility

Verified against the front code: `?timezone=` and the `Authorization` header are added to every agent request (since ≥2021, with a `UTC` fallback), and the pre-auth 401 on the first capabilities call is the exact behaviour of agent-nodejs in production today — the front handles it by refetching on `authenticationSucceeded`. Expect a small uptick of "Agent logout: 401" warnings on Ruby agents after release, same as Node.

## Regression specs

- Capabilities / datasource charts / scope invalidation: 401 with no `Authorization` header, permissions never instantiated, `Whitelist.check_ip` invoked when the request carries `action_dispatch.remote_ip`.
- Mutation-tested: reverting the fix, dropping only the `build(args)` call, or keeping the call with the wrong base class each make the new specs fail.
- CallerParser: wrong-signature and expired JWTs → `UnauthorizedError`.

## Out of scope, tracked separately

- [PRD-914](https://linear.app/forestadmin/issue/PRD-914/agent-ruby-ip-whitelist-rejection-raises-a-module-instead-of-an): the whitelist rejection path raises a module (`Net::HTTPExceptions`) → `TypeError` → 500 instead of Node's 403 (fail-closed by accident, pre-existing on all routes).
- The proof spec on `feature/prd-904-security-audit-proof-suite` asserts the old unauthenticated behaviour and will need flipping when that branch lands.
- ~~Untestable locally: the MCP server token (`describe-collection`) against a Ruby agent~~ **Validated**: minted a token through the real compiled MCP-server code (`AuthService#getUserInfo` mapping — `team = teams[0]` is unconditional — plus `ForestOAuthProvider.toSnakeCaseKeys` and the verbatim payload construction, Forest server stubbed locally) and ran it through the real `CallerParser` and the authenticated capabilities route on this branch: caller parses with `team` intact, route answers 200, and the same request without the token gets 401. Ruby's `Caller.new` also swallows the MCP token's extra claims (`serverToken`, `scopes`, camelCase duplicates) via `**_extra_args`.

## Test plan

- [x] `forest_admin_agent`: 796 examples, 0 failures
- [x] `forest_admin_rails`: 110 examples, 0 failures (now also in CI)
- [x] `rubocop`: 802 files, no offenses

> [!NOTE]
> ### Enforce authentication and IP whitelist on capabilities, charts, and scope invalidation routes
> - Changes `Collections`, `ApiChartDatasource`, and `ScopeInvalidation` routes to extend `AbstractAuthenticatedRoute` and call `build(args)` in their handlers, applying authentication and IP whitelist checks that were previously skipped.
> - Adds a `JWT::DecodeError` rescue in [`CallerParser.decode_token`](https://github.com/ForestAdmin/agent-ruby/pull/349/files#diff-83e1f346dcd1d417b18f30cdf95cc3d0bb5f430845a5df7f42e504cd17940a1f) to raise a consistent `UnauthorizedError` with the message `'Invalid or expired authentication token.'` instead of propagating raw JWT exceptions.
> - Adds `forest_admin_rails` to the CI test matrix and updates coverage reporting to include it.
> - Behavioral Change: Requests to the capabilities, datasource chart, and scope invalidation endpoints without a valid token or from a non-whitelisted IP will now receive 401/403 responses rather than proceeding.
>
> 
> #### Changes since #349 opened
>
> - Refactored `ForestAdminAgent::Routes::Capabilities::Collections.handle_request` to retrieve datasource from context object [0cefea7]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1e1e75c.</sup>
> 
>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,6 +71,7 @@ jobs:
           - forest_admin_datasource_active_record
           - forest_admin_datasource_customizer
           - forest_admin_datasource_toolkit
+          - forest_admin_rails
           - forest_admin_datasource_mongoid
           - forest_admin_rpc_agent
           - forest_admin_datasource_rpc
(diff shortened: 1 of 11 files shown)


.github/workflows/build.yml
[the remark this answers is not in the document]
PMerlet: Fixed in 1e1e75cd — coverage job matrix moved to `ruby-version: ["4.0"]` to match the artifact names.

One precision on the impact: the step did not actually fail — the action runs with `skip-errors: true`, so it logged `##[warning] No code coverage data files were found` and sent **zero** coverage. Checked the logs of the previous run on this PR: the download step fetched `4.0-*` artifacts while every configured file path pointed at `3.4-*`, for **all** packages, not just the newly added forest_admin_rails. So repo-wide coverage reporting has been silently broken since the artifact upload moved to ruby 4.0; this fix restores it for every package.

packages/forest_admin_agent/lib/forest_admin_agent/routes/capabilities/collections.rb
```diff
@@ -16,6 +16,7 @@ def setup_routes
         end
 
         def handle_request(args = {})
+          build(args)
           datasource = ForestAdminAgent::Facades::Container.datasource
```
bexchauveto: `build(args)`'s return value is discarded here, and the datasource is still fetched via `Facades::Container.datasource`. The sibling routes touched in this same PR (`api_chart_datasource.rb`'s `handle_api_chart`/`handle_smart_chart`) were updated to `context = build(args)` + `context.datasource`/`context.caller`. Suggest matching that for consistency.

```suggestion
          context = build(args)
          datasource = context.datasource
```

PMerlet: Good catch, applied in 0cefea72 — `context = build(args)` + `context.datasource`, same pattern as the chart routes. No behaviour change (`AbstractRoute#build` sets `context.datasource` from the same facade), specs and rubocop green.

bexchauveto (approved):

PR #350, opened 2026-08-10
fix(agent): scope and authorize related write routes on the mutated collection

PMerlet:
## Summary

Write-side counterpart of #348 (PRD-897), found during its adversarial review. The relation **write** routes intersected the **parent** collection's scope, unnested, into filters that mutate the **child** collection — and in two places skipped the permission check on the collection actually being updated. All changes mirror `agent-nodejs` (`associate-related.ts`, `update-relation.ts`, `dissociate-delete-related.ts`).

- **Associate (OneToMany / PolymorphicOneToMany)**: the child update filter now intersects `get_scope(child_collection)` instead of the parent's scope. The `edit` check stays on the parent, matching Node.
- **Update relation (OneToOne / PolymorphicOneToOne)**: the route-level `can?(:edit, parent)` moves into the branches; these check `can?(:edit, child_collection)` and use the child's scope in the break-old / create-new filters (aggregate included). Previously a role with `edit` on the parent but not the child could rewrite the child's foreign keys.
- **Update relation (ManyToOne / PolymorphicManyToOne)**: the parent fk-overwrite filter had no scope at all; it now intersects `get_scope(parent)`, so a caller can no longer re-point a parent record their scope excludes.
- **Dissociate (non-delete mode)**: checks `can?(:edit, child_collection)` — the child is the collection whose origin key is nulled. Delete mode was already correct.

Not included (parity with Node): the ManyToMany associate branch creates the through record with no scope on either agent; a negative `get_scope` assertion pins that behavior.

## Test plan

- Specs updated with the hardened pattern from #348: each test stubs a real scope tree, asserts it lands in the right filter, and adds positive/negative `get_scope` / `can?` assertions per collection — a regression re-intersecting the parent scope or checking the wrong collection fails the suite.
- `bundle exec rspec` (forest_admin_agent): 788 examples, 0 failures.
- `bundle exec rubocop` on changed files: no offenses.

PRD-908

> [!NOTE]
> ### Fix permission scope checks on related write routes to use the mutated collection
> - In `AssociateRelated`, scope filters for one-to-many and polymorphic one-to-many now use `context.child_collection` instead of `context.collection`.
> - In `DissociateRelated`, the edit permission check in non-delete mode now targets `context.child_collection` instead of the parent.
> - In `UpdateRelated`, permission checks are moved from a single entry-point call into each relation-specific helper, applying `context.child_collection` for OneToOne/polymorphic-OneToOne writes and `context.collection` for ManyToOne writes.
> - Scope filters passed to `ConditionTree` intersections now consistently reference the collection being mutated, not the collection the route was entered through.
> - Risk: Any code relying on parent-collection scopes being applied to child-collection writes will now enforce stricter, relation-correct permissions.
>
> 
> #### Changes since #350 opened
>
> - Moved authorization check to occur before ID parsing in `ForestAdminAgent::Routes::Resources::Related::UpdateRelated.handle_request` and added `mutated_collection` helper method that returns `context.child_collection` for `OneToOne` and `PolymorphicOneToOne` relations or `context.collection` otherwise [72a09cc]
> - Removed inline authorization checks from `update_many_to_one`, `update_polymorphic_many_to_one`, `update_one_to_one`, and `update_polymorphic_one_to_one` methods in `ForestAdminAgent::Routes::Resources::Related::UpdateRelated` [72a09cc]
> - Extracted filter construction logic into `scoped_fk_owner_filter` helper method in `ForestAdminAgent::Routes::Resources::Related::UpdateRelated` and updated `update_many_to_one` and `update_polymorphic_many_to_one` to use it [72a09cc]
> - Added test coverage verifying authorization occurs before ID parsing in `ForestAdminAgent::Routes::Resources::Related::UpdateRelated` [72a09cc]
> - Updated inline comment documentation in `ForestAdminAgent::Routes::Resources::Related::UpdateRelated.handle_request` method [1884010]
> 
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 385aa62. 3 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/related/update_related.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/related/update_related.rb
@@ -22,7 +22,6 @@ module ForestAdminAgent
 
           def handle_request(args = {})
             context = build(args)
-            context.permissions.can?(:edit, context.collection)
 
             relation = context.collection.schema[:fields][args[:params]['relation_name']]
             parent_primary_key_values = Utils::Id.unpack_id(context.collection, args[:params]['id'])
@@ -48,16 +47,27 @@ module ForestAdminAgent
           private
 
           def update_many_to_one(relation, parent_primary_key_values, linked_primary_key_values, context)
+            context.permissions.can?(:edit, context.collection)
+
             foreign_value = if linked_primary_key_values
                               Collection.get_value(context.child_collection, context.caller, linked_primary_key_values,
                                                    relation.foreign_key_target)
                             end
             fk_owner = ConditionTree::ConditionTreeFactory.match_ids(context.collection, [parent_primary_key_values])
-            context.collection.update(context.caller, Filter.new(condition_tree: fk_owner),
-                                      { relation.foreign_key => foreign_value })
+            filter = Filter.new(
+              condition_tree: ConditionTree::ConditionTreeFactory.intersect(
+                [
+                  context.permissions.get_scope(context.collection),
+                  fk_owner
+                ]
+              )
+            )
+            context.collection.update(context.caller, filter, { relation.foreign_key => foreign_value })
           end
 
           def update_polymorphic_many_to_one(relation, parent_primary_key_values, linked_primary_key_values, context)
+            context.permissions.can?(:edit, context.collection)
+
             foreign_value = if linked_primary_key_values
                               Collection.get_value(
                                 context.child_collection,
@@ -69,9 +79,17 @@ module ForestAdminAgent
 
             polymorphic_type = context.child_collection.name.gsub('__', '::')
             fk_owner = ConditionTree::ConditionTreeFactory.match_ids(context.collection, [parent_primary_key_values])
+            filter = Filter.new(
+              condition_tree: ConditionTree::ConditionTreeFactory.intersect(
+                [
+                  context.permissions.get_scope(context.collection),
+                  fk_owner
+                ]
+              )
+            )
             context.collection.update(
               context.caller,
-              Filter.new(condition_tree: fk_owner),
+              filter,
               {
                 relation.foreign_key => foreign_value,
                 relation.foreign_key_type_field => polymorphic_type
@@ -80,6 +98,8 @@ module ForestAdminAgent
           end
 
           def update_polymorphic_one_to_one(relation, parent_primary_key_values, linked_primary_key_values, context)
+            context.permissions.can?(:edit, context.child_collection)
+
             origin_value = Collection.get_value(context.collection, context.caller, parent_primary_key_values,
                                                 relation.origin_key_target)
 
@@ -88,6 +108,8 @@ module ForestAdminAgent
           end
 
           def update_one_to_one(relation, parent_primary_key_values, linked_primary_key_values, context)
+            context.permissions.can?(:edit, context.child_collection)
+
             origin_value = Collection.get_value(context.collection, context.caller, parent_primary_key_values,
                                                 relation.origin_key_target)
 
@@ -101,7 +123,7 @@ module ForestAdminAgent
             old_fk_owner_filter = Filter.new(
               condition_tree: ConditionTree::ConditionTreeFactory.intersect(
                 [
-                  context.permissions.get_scope(context.collection),
+                  context.permissions.get_scope(context.child_collection),
                   ConditionTree::Nodes::ConditionTreeBranch.new(
                     'And',
                     [
@@ -150,7 +172,7 @@ module ForestAdminAgent
               Filter.new(
                 condition_tree: ConditionTree::ConditionTreeFactory.intersect(
                   [
-                    context.permissions.get_scope(context.collection), new_fk_owner
+                    context.permissions.get_scope(context.child_collection), new_fk_owner
                   ]
                 )
               ),
@@ -164,7 +186,7 @@ module ForestAdminAgent
             old_fk_owner_filter = Filter.new(
               condition_tree: ConditionTree::ConditionTreeFactory.intersect(
                 [
-                  context.permissions.get_scope(context.collection),
+                  context.permissions.get_scope(context.child_collection),
                   ConditionTree::Nodes::ConditionTreeLeaf.new(
                     relation.origin_key,
                     ConditionTree::Operators::EQUAL,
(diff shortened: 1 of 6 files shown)


packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/related/update_related.rb
```diff
@@ -101,7 +123,7 @@ def break_old_polymorphic_one_to_one_relationship(relation, origin_value, linked
             old_fk_owner_filter = Filter.new(
               condition_tree: ConditionTree::ConditionTreeFactory.intersect(
                 [
-                  context.permissions.get_scope(context.collection),
+                  context.permissions.get_scope(context.child_collection),
```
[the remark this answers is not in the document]
PMerlet: The suggested change would reintroduce the exact bug class this PR (and #348 on the read side) fixes: `get_scope(context.collection)` returns a condition tree expressed on the **parent's** fields, and intersecting it unnested into a filter executed against the **child** collection either errors (field absent on the child) or silently filters on the wrong column when names collide (`status`, `company_id`, …). It also drops the child's own row-level restriction, letting a caller null out child rows their scope excludes.

Using the child's scope here is deliberate parity with `agent-nodejs`: `update-relation.ts:96` fetches `getScope(this.foreignCollection)` and passes it to `breakOldOneToOneRelationship`, which intersects it into the same aggregate + update filters (`update-relation.ts:142-167`).

The scenario you describe (old child out of scope → break is a no-op, create still links the new child, leaving two rows pointing at the parent) is real, but it exists identically in agent-nodejs and follows from scope semantics: a scope restricts which rows the caller may mutate, and silently updating an out-of-scope row would be a scope bypass — strictly worse. Actually *blocking* the replacement would require an explicit pre-check raising Forbidden when an out-of-scope child holds the fk, which is a behavior change to make in both agents together, not a scope swap in this filter. Keeping as is.

packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/related/update_related.rb
```diff
@@ -22,7 +22,6 @@ def setup_routes
 
           def handle_request(args = {})
             context = build(args)
-            context.permissions.can?(:edit, context.collection)
 
             relation = context.collection.schema[:fields][args[:params]['relation_name']]
```
bexchauveto, line 26: The top-level `context.permissions.can?(:edit, context.collection)` check that used to run first in `handle_request` was moved down into each relation-type handler (`update_many_to_one`/`update_polymorphic_many_to_one`/`update_one_to_one`/`update_polymorphic_one_to_one`), which are only reached after `relation` lookup (line 26) and both `Utils::Id.unpack_id` calls (lines 27 and 30).

That means a caller with no edit permission who sends a malformed `id` or malformed `data.id` now gets a PK-shape validation error (e.g. "Expected 2 primary keys, found 1") instead of a 403 — their input is parsed and can raise before authorization is ever checked. Not something PRD-908 asked for; looks like a side effect of pushing the check into the per-relation handlers. Worth keeping an early can?/authorization gate in `handle_request` before the unpacking, in addition to (or instead of) the per-branch checks that pick parent vs. child collection.

packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/related/update_related.rb
```diff
@@ -48,16 +47,27 @@ def handle_request(args = {})
...
-            context.collection.update(context.caller, Filter.new(condition_tree: fk_owner),
-                                      { relation.foreign_key => foreign_value })
+            filter = Filter.new(
+              condition_tree: ConditionTree::ConditionTreeFactory.intersect(
+                [
+                  context.permissions.get_scope(context.collection),
+                  fk_owner
+                ]
+              )
+            )
```
bexchauveto: This scope-intersected owner `Filter` block is byte-for-byte duplicated in `update_polymorphic_many_to_one` (lines 82-89), aside from variable naming. Minor, but a future change to how the owner filter is scoped needs to land in both places, and it's easy to update one and forget the other, leaving the two ManyToOne update paths silently inconsistent. Worth extracting into a small shared helper.

packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/related/update_related.rb
PMerlet: Good catch — fixed in 72a09cc5. The `can?(:edit, …)` check now runs in `handle_request` before both `unpack_id` calls, picking the collection the relation type mutates (child for OneToOne/PolymorphicOneToOne, parent otherwise), so an unauthorized caller gets a 403 before their input is parsed. Added a spec that pins the ordering (`can?` raising Forbidden + `unpack_id` never received).

packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/related/update_related.rb
PMerlet: Done in 72a09cc5 — extracted `scoped_fk_owner_filter`, used by both the ManyToOne and PolymorphicManyToOne branches.

packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/related/update_related.rb
```diff
@@ -22,9 +22,14 @@ def setup_routes
...
           def handle_request(args = {})
             context = build(args)
-            context.permissions.can?(:edit, context.collection)
 
             relation = context.collection.schema[:fields][args[:params]['relation_name']]
+
+            # Authorize on the collection the relation type mutates, before any
+            # id parsing, so unauthorized callers get a 403 rather than a
+            # validation error on their input.
+            context.permissions.can?(:edit, mutated_collection(relation, context))
```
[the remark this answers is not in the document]
PMerlet: Pre-existing behavior, not introduced by this PR: an unknown `relation_name` never reaches `mutated_collection`. `build(args)` — the first statement of `handle_request` — already dereferences `relation.type` in `AbstractRelatedRoute#build` (abstract_related_route.rb:8), so on `main` the same `NoMethodError` was raised there, before the old top-level `can?` check too. The ordering between "unknown relation error" and the permission check is unchanged by this PR.

Making that a clean error instead of a `NoMethodError` is worth doing, but it belongs in `AbstractRelatedRoute#build` (it affects all six related routes, read and write) — a separate change, mirroring `SchemaUtils.getRelation`'s validation error in agent-nodejs, which also runs before the permission checks there.

bexchauveto (approved):

PR #352, opened 2026-08-11
feat(agent): accept the Forest-Projection header on record routes

PMerlet:
## What

Ruby counterpart of agent-nodejs#1813 (get-one) and agent-nodejs#1822 (list, relationship, csv). Until now the Ruby agent had **no** `Forest-Projection` support at all, and did not announce the capability.

The get-one, list, relationship list and csv export routes now read the projection from the `Forest-Projection` header (format `field,relation:subField`), taking precedence over the `fields[...]` query params. Relationship and list URLs observed in production routinely approach the 2KB AWS WAF `SizeRestrictions_QUERYSTRING` default, and header-based projections are a prerequisite for projecting workspaces.

## How

`QueryStringParser.parse_projection_from_header` reads `HTTP_FOREST_PROJECTION` (the Rails controller already forwards `request.headers.to_h`), and `parse_projection_from_request` implements the header-then-query-string precedence.

Four of the five routes already went through `parse_projection_with_pks`, so they pick the header up for free. The two csv routes deliberately do not add primary keys — the exported columns must match the `header` label row — so both call `parse_projection_from_request` directly.

The header fails closed: an invalid one raises `Invalid Forest-Projection header:` instead of falling back to the query string, and empty segments (`,` `a,` `a,,b` `a:` `:b`) are rejected rather than producing an empty projection.

## Polymorphism

The header path reproduces the query-string semantics rather than bypassing them:

- a `PolymorphicManyToOne` asked bare, or with an explicit `:*`, is projected as `relation:*`;
- a nested field asked under it (`addressable:email`) is rejected by `FieldValidator` instead of being silently widened;
- the discriminant type field is added whenever the polymorphic relation or its foreign key is requested, without duplicating it if the caller already asked for it.

## Capabilities

The agent now announces `canUseProjectionViaHeader` and `canUseProjectionViaHeaderOnList`, matching the Node agent so the frontend can gate on the same flags across both stacks. Query-string parsing and `canUseMultipleFieldsProjectionOnRelation` are untouched, so older frontends keep working unchanged.

## Notes

- CORS: nothing to do, the Rails engine echoes `Access-Control-Request-Headers`.
- The Rails controller is the only entry point building route args, so the header reaches every deployment. The RPC agent passes no `:headers` key and takes its projection from the body; it is unaffected.
- `csv_related` no longer adds primary keys, aligning it with `csv` and with agent-nodejs: the extra column had no matching label in the header row.
- The spec suite has a pre-existing, unrelated flakiness (a filesystem race in `spec_helper`'s cache cleanup) that can redden arbitrary examples.

fixes PRD-928

### DIFF
--- a/packages/forest_admin_agent/lib/forest_admin_agent/routes/capabilities/collections.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/routes/capabilities/collections.rb
@@ -65,6 +65,8 @@ module ForestAdminAgent
               nativeQueryConnections: connections,
               agentCapabilities: {
                 canUseProjectionOnGetOne: true,
+                canUseProjectionViaHeader: true,
+                canUseProjectionViaHeaderOnList: true,
                 canUseMultipleFieldsProjectionOnRelation: true
               }
             },
--- a/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/csv.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/csv.rb
@@ -37,7 +37,7 @@ module ForestAdminAgent
             sort: QueryStringParser.parse_sort(context.collection, args),
             segment: QueryStringParser.parse_segment(context.collection, args)
           )
-          projection = QueryStringParser.parse_projection(context.collection, args)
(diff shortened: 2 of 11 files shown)


christophebrun-forest (approved):
Nice, LGTM

PR #353, opened 2026-08-12
fix(agent): authorize create with a linked one-to-one on the foreign collection

PMerlet:
Fixes [PRD-921](https://linear.app/forestadmin/issue/PRD-921/agent-ruby-create-with-a-linked-one-to-one-writes-the-foreign). Ruby counterpart of PRD-916, fixed in agent-nodejs [#1819](https://github.com/ForestAdmin/agent-nodejs/pull/1819).

## The bug

`POST /forest/:collection` with a linked `OneToOne` / `PolymorphicOneToOne` relation asserted only `add` on the **parent** collection, then updated the **foreign** collection with no authorization and no scope:

```ruby
condition_tree = ConditionTreeFactory.match_records(foreign_collection, [primary_key_values])
foreign_collection.update(context.caller, Filter.new(condition_tree: condition_tree), patch)
```

A user with `add` on `books` but no `edit` on `passports` could rewrite the foreign key of an arbitrary passport — stealing it from its owner — including a passport outside their scope. Worse than the Node instance on two counts: Node at least asserted `edit` (on the wrong side) and kept the writes scope-bounded.

## The fix

- Assert `edit` on the **foreign** collection, once per linked one-to-one relation, and **before** the parent record is created — a denied link writes nothing at all, and a denial on one relation no longer lets writes on the others through. The parent-side `add` check is unchanged.
- Intersect `get_scope(foreign_collection)` into the update filter, mirroring `create_new_one_to_one_relationship` in `update_related.rb` (PRD-908, #350).
- Resolve **everything** the linking pass needs — the `edit` assertion, the scope, and the unpacked linked id — in a single pre-create pass, and carry it in the relation hash. `get_scope` reads the `forest.rendering` cache that nothing earlier in this route warms, and unlike `can?` it has no `permission_system?` bypass; leaving it below `create` meant a cold cache, a Forest API 5xx or an unresolvable team/user raised **after** the parent row was committed, so the 500 the caller retried duplicated the parent. Same shape for a malformed packed id. The "a rejected link writes nothing at all" invariant now holds for failed lookups, not only for denied permissions.
- Single `linked_one_to_one_relations` helper feeding both the permission pass and the linking pass, so widening the guard on one side can no longer silently drop the check on the other.

## Observable contract when the scope excludes the target

A caller linking a foreign record outside their scope gets a **201 with the parent created and the relation absent** — the intersected filter matches no row and `entity&.update!` is a no-op, indistinguishable from linking an id that does not exist. This is deliberate: it is what agent-nodejs [#1819](https://github.com/ForestAdmin/agent-nodejs/pull/1819) and `create_new_one_to_one_relationship` in `update_related.rb` both do, and the PRD prescribes the mechanism without asking for a 403. Counting first to raise `ForbiddenError` would diverge from Node and cost an extra query; the contract is instead pinned by a test and stated at the call site.

Two defects on the rewritten lines, fixed along the way and each pinned by a test — flagging them explicitly as they sit outside the strict scope of the PRD:

- the linked id was unpacked against **`context.collection`** (the parent) while the `match_records` that follows targets the foreign collection, so the filter carried the wrong primary key names as soon as the two sides disagreed;
- a one-to-one relationship carrying `data: null` raised a `NoMethodError` (`value['data']['id']`); it is now ignored, matching Node's `linked !== null`.

## Deliberate divergence from Node

Node runs two updates (null the old foreign-key owner, then claim the new one); this route keeps its single update. On a create, `origin_value` is the id of a record that has just been born, so no foreign record can already carry it — Node's first update is a no-op costing one query. The PRD notes the same, and it has no bearing on the authorization gap.

## Tests

`store_spec.rb`: scope intersected (one-to-one and polymorphic), a denied `edit` creating neither parent nor link, `403` raised before the linked id is parsed (the invariant PRD-908 established on `update_related.rb`), a failed scope lookup and a malformed linked id each creating neither parent nor link, the silent no-op when the scope excludes the target, `data: null` ignored, and the parent-side `add` check pinned so removing it would not go unnoticed.

The polymorphic fixture carries `origin_type_value: 'Person::Legacy'` on a collection named `person`, so the `addressable_type` assertion pins `context.collection.name` as the source of the type column instead of passing under either candidate.

Permission assertions go through a collected `[action, collection_name]` list rather than `having_attributes` on the collection instances: on failure, rspec inspects the decorator chain and the run hangs instead of reporting.

Every new assertion was checked against the faulty implementation it guards — the pre-fix route, `unpack_id` hoisted back above the permission check, the parent collection restored as the unpack target, `get_scope` and `unpack_id` dropped back below `create`, and `schema.origin_type_value` swapped in for the collection name — and each mutation is caught.

842 examples, 0 failures on `forest_admin_agent`; rubocop clean.

> [!NOTE]
> ### Authorize edit on linked one-to-one foreign collections before creating a record
> - The `Store#handle_request` action now resolves linked one-to-one (and polymorphic one-to-one) relations from the request payload before creating the main record, and calls `can?(:edit)` on each foreign collection, aborting if any permission is denied.
> - `link_one_to_one_relations` is refactored to accept a precomputed relations array and now unpacks the foreign record's primary key against the foreign collection (not the current collection), fixing incorrect record matching.
> - Scope from `permissions.get_scope` is intersected into the update filter for each foreign collection, ensuring scoped access controls are respected during relation linking.
> - One-to-one relationships with `nil` data ids are skipped rather than processed.
>
> 
> #### Changes since #353 opened
>
> - Added authorization checks for one-to-one linked relations during record creation [7a1a75c]
> - Removed inline comments from test files [7a1a75c]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 222e53d.</sup>
> 
>

### DIFF
--- a/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/store.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/store.rb
@@ -17,7 +17,7 @@ module ForestAdminAgent
         def handle_request(args = {})
           context = build(args)
           context.permissions.can?(:add, context.collection)
-          relations = resolve_linked_one_to_one_relations(args, context)
+          relations = authorized_linked_one_to_one_relations(args, context)
           data = format_attributes(args, context.collection)
           record = context.collection.create(context.caller, data)
           link_one_to_one_relations(relations, record, context)
@@ -40,14 +40,10 @@ module ForestAdminAgent
 
         private
 
-        # Everything the linking pass needs is authorized and resolved before the parent record is
-        # created: a denied edit, an unresolvable scope or a malformed linked id must leave no
-        # committed parent row behind.
-        def resolve_linked_one_to_one_relations(args, context)
+        def authorized_linked_one_to_one_relations(args, context)
           linked_one_to_one_relations(args, context).map do |relation|
             foreign_collection = relation[:foreign_collection]
 
-            # Must run before unpack_id: unauthorized callers get a 403, not a validation error
             context.permissions.can?(:edit, foreign_collection)
 
             relation.merge(
@@ -77,8 +73,6 @@ module ForestAdminAgent
           }
         end
 
-        # A foreign record excluded by the caller's scope matches no row, so the link is silently
-        # dropped, as on the update_related route and in agent-nodejs.
         def link_one_to_one_relations(relations, record, context)
           relations.each do |relation|
             schema = relation[:schema]
--- a/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/store.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/store.rb
@@ -17,9 +17,11 @@ module ForestAdminAgent
         def handle_request(args = {})
           context = build(args)
           context.permissions.can?(:add, context.collection)
+          relations = linked_one_to_one_relations(args, context)
+          assert_can_edit_linked_collections(relations, context)
           data = format_attributes(args, context.collection)
           record = context.collection.create(context.caller, data)
-          link_one_to_one_relations(args, record, context)
+          link_one_to_one_relations(relations, record, context)
           id = ForestAdminDatasourceToolkit::Utils::Record.primary_keys(context.collection, record)
           filter = ForestAdminDatasourceToolkit::Components::Query::Filter.new(
             condition_tree: ConditionTree::ConditionTreeFactory.match_ids(context.collection, [id])
@@ -37,24 +39,49 @@ module ForestAdminAgent
           }
         end
 
-        def link_one_to_one_relations(args, record, context)
-          args[:params][:data][:relationships]&.map do |field, value|
+        private
+
+        def assert_can_edit_linked_collections(relations, context)
+          relations.each { |relation| context.permissions.can?(:edit, relation[:foreign_collection]) }
+        end
+
+        def linked_one_to_one_relations(args, context)
+          args[:params][:data][:relationships]&.filter_map do |field, value|
             schema = context.collection.schema[:fields][field]
             next unless %w[OneToOne PolymorphicOneToOne].include?(schema.type)
 
-            primary_key_values = Utils::Id.unpack_id(context.collection, value['data']['id'], with_key: true)
-            foreign_collection = context.datasource.get_collection(schema.foreign_collection)
-            # Load the value that will be used as origin_key
+            id = value.dig('data', 'id')
+            next if id.nil?
+
+            {
+              schema: schema,
+              foreign_collection: context.datasource.get_collection(schema.foreign_collection),
+              id: id
+            }
+          end || []
+        end
+
+        def link_one_to_one_relations(relations, record, context)
+          relations.each do |relation|
+            schema = relation[:schema]
+            foreign_collection = relation[:foreign_collection]
+            primary_key_values = Utils::Id.unpack_id(foreign_collection, relation[:id], with_key: true)
             origin_value = record[schema.origin_key_target]
 
-            # update new relation (may update zero or one records).
             patch = { schema.origin_key => origin_value }
             if schema.type == 'PolymorphicOneToOne'
               patch[schema.origin_type_field] =
                 context.collection.name.gsub('__', '::')
             end
-            condition_tree = ConditionTree::ConditionTreeFactory.match_records(foreign_collection, [primary_key_values])
-            filter = Filter.new(condition_tree: condition_tree)
+            new_fk_owner = ConditionTree::ConditionTreeFactory.match_records(foreign_collection, [primary_key_values])
+            filter = Filter.new(
+              condition_tree: ConditionTree::ConditionTreeFactory.intersect(
+                [
+                  context.permissions.get_scope(foreign_collection),
+                  new_fk_owner
+                ]
+              )
+            )
             foreign_collection.update(context.caller, filter, patch)
           end
         end
--- a/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/store.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/store.rb
@@ -17,8 +17,7 @@ module ForestAdminAgent
         def handle_request(args = {})
           context = build(args)
           context.permissions.can?(:add, context.collection)
-          relations = linked_one_to_one_relations(args, context)
-          assert_can_edit_linked_collections(relations, context)
+          relations = resolve_linked_one_to_one_relations(args, context)
           data = format_attributes(args, context.collection)
           record = context.collection.create(context.caller, data)
           link_one_to_one_relations(relations, record, context)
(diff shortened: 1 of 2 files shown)


packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/store.rb
```diff
@@ -17,9 +17,11 @@ def setup_routes
         def handle_request(args = {})
           context = build(args)
           context.permissions.can?(:add, context.collection)
+          relations = linked_one_to_one_relations(args, context)
+          assert_can_edit_linked_collections(relations, context)
           data = format_attributes(args, context.collection)
           record = context.collection.create(context.caller, data)
-          link_one_to_one_relations(args, record, context)
+          link_one_to_one_relations(relations, record, context)
```
hercemer42, line 23: Claude Opus 5 (claude-opus-5): **Should fix**

**Applies to:** `store.rb:80` (`context.permissions.get_scope(foreign_collection)`), reached from this line.

A Forest API hiccup on a create with a linked one-to-one leaves a committed parent row and a 500, and the retry duplicates it. The invariant this PR establishes for a *denied* link — nothing is written at all — does not hold for a *failed* permission lookup, because the second new permissions call stayed below `create`.

`can?` warms `forest.has_permission` / `forest.users` / `forest.collections` (`services/permissions.rb:263, 169, 206`). `get_scope` reads a different key, `forest.rendering` (`:248`), which nothing earlier in this route touches — so on a cold or expired rendering cache the request's first `/liana/v4/permissions/renderings/:id` fetch happens *after* `context.collection.create`, and it raises on a 5xx (`:379-385`) or when team/user resolve to nil (`:151-155`). `get_scope` also has no `permission_system?` bypass, unlike `can?` (`:39`), so installations with the permission system disabled newly make that call from the create path.

Resolving the scope in the same pre-create pass as the `edit` assertion and carrying it in the relation hash closes it. `Utils::Id.unpack_id` at `:68` has the same shape — a malformed packed id still 500s after the parent exists — and moving it into `linked_one_to_one_relations` closes both for free.

packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/store.rb
```diff
@@ -37,24 +39,49 @@ def handle_request(args = {})
...
               patch[schema.origin_type_field] =
                 context.collection.name.gsub('__', '::')
             end
-            condition_tree = ConditionTree::ConditionTreeFactory.match_records(foreign_collection, [primary_key_values])
-            filter = Filter.new(condition_tree: condition_tree)
+            new_fk_owner = ConditionTree::ConditionTreeFactory.match_records(foreign_collection, [primary_key_values])
+            filter = Filter.new(
+              condition_tree: ConditionTree::ConditionTreeFactory.intersect(
+                [
+                  context.permissions.get_scope(foreign_collection),
```
hercemer42: Claude Opus 5 (claude-opus-5): **Should fix**

**Applies to:** the PR description, and `routes/resources/related/update_related.rb:193-207` (not in this diff).

`get_scope(foreign_collection)` is the right argument — it matches agent-nodejs #1819, which intersects `getScope(foreignCollection)`. But the description says this mirrors `create_new_one_to_one_relationship` in `update_related.rb`, and that method intersects `get_scope(context.collection)` — the *parent's* scope — into a `context.child_collection.update` (`update_related.rb:201`; same at `:104`, `:153`, `:167`).

Consequence on that route, which PRD-921 cites as the correct reference: for a team carrying a scope on the parent collection, `PUT /forest/:collection/:id/relationships/:relation` filters the foreign collection by parent field names — a 500 when the field is absent there, wrong rows when it happens to exist — while a scope on the *foreign* collection is never applied, so the hole PRD-908 set out to close is still open on the one-to-one paths. PRD-908's own note that Node uses `getScope(this.foreignCollection)` is what the Ruby side missed.

Two asks: drop the "mirroring `update_related.rb`" claim from the description, since this PR deliberately (and correctly) diverges; and open a follow-up for those four parent-scope call sites.

packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/store.rb
```diff
@@ -37,24 +39,49 @@ def handle_request(args = {})
...
             patch = { schema.origin_key => origin_value }
             if schema.type == 'PolymorphicOneToOne'
               patch[schema.origin_type_field] =
                 context.collection.name.gsub('__', '::')
             end
-            condition_tree = ConditionTree::ConditionTreeFactory.match_records(foreign_collection, [primary_key_values])
-            filter = Filter.new(condition_tree: condition_tree)
+            new_fk_owner = ConditionTree::ConditionTreeFactory.match_records(foreign_collection, [primary_key_values])
+            filter = Filter.new(
+              condition_tree: ConditionTree::ConditionTreeFactory.intersect(
```
hercemer42, line 91: Claude Opus 5 (claude-opus-5): **Preferential**

**Applies to:** `store.rb:84` (`foreign_collection.update`).

Newly reachable state: a caller linking a foreign record outside their scope gets a **201 with the parent created and the relation silently absent** — indistinguishable from linking an id that does not exist. Before this PR the write went through (the bug being fixed); now the intersected filter matches zero rows and nothing reports it. `datasource_active_record/collection.rb:44-49` is `…build.first` then `entity&.update!(data)`, so a nil entity is a no-op, not an error.

This matches Node #1819 and `update_related.rb`, and PRD-921 prescribes the mechanism without stating the observable outcome — so it reads as deliberate silence rather than a regression. Worth settling explicitly all the same: either count first and raise `ForbiddenError` (`update_related.rb:149-151` already has that shape), or state the silent-no-op contract in the PR body. Either way the excluded-target case deserves a test — both new scope tests assert filter shape only, never a scope that actually excludes the target.

packages/forest_admin_agent/spec/lib/forest_admin_agent/routes/resources/store_spec.rb
```diff
@@ -245,6 +261,147 @@ def respond_to?(arg)
...
+                }
+              )
+              @datasource.add_collection(collection_address)
+              @datasource.get_collection('person').schema[:fields]['address'] =
+                Relations::PolymorphicOneToOneSchema.new(
+                  origin_key: 'addressable_id',
+                  origin_key_target: 'id',
+                  foreign_collection: 'address',
+                  origin_type_field: 'addressable_type',
+                  origin_type_value: 'person'
```
hercemer42: Claude Opus 5 (claude-opus-5): **Preferential**

`origin_type_value: 'person'` on a collection named `person` makes the `'addressable_type' => 'person'` assertion at `:393` pass under both candidate implementations — `context.collection.name.gsub('__', '::')`, which is what the code does, and `schema.origin_type_value`, which is what the schema carries. The new polymorphic test pins every part of that write except which source of truth feeds the type column, so a change there lands green.

`origin_type_value: 'Person::Legacy'`, collection still named `person`, states the intent.

hercemer42 (changes requested):

packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/store.rb
PMerlet: Fixed in 53da1410 — the diagnosis holds. `get_scope` reads `forest.rendering` (`services/permissions.rb:248`), which nothing earlier in this route warms, and unlike `can?` it has no `permission_system?` bypass — so the request's first rendering fetch happened after `context.collection.create`, and a 5xx or a nil team/user raised on a committed parent, whose retry duplicated it.

The `edit` assertion, the scope and `Utils::Id.unpack_id` are now resolved in one pre-create pass (`resolve_linked_one_to_one_relations`) and carried in the relation hash, so nothing below `create` can raise on a permission or id lookup. Two tests pin it — a failed scope lookup and a malformed packed id each creating neither parent nor link — and both were checked against the previous ordering. Splitting the relation lookup in two also clears the qlty complexity flag.

packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/store.rb
PMerlet: Kept as-is, and documented. The silent 201-with-no-relation is now stated in the PR body under *Observable contract when the scope excludes the target*, spelled out at the call site, and pinned by a test (`drops the link silently when the scope excludes the linked record`) that asserts the excluding scope reaches the filter and that the create still succeeds — the gap you flagged: the two existing scope tests only assert filter shape.

Chose the no-op over counting-then-`ForbiddenError` to stay aligned with agent-nodejs #1819 and `create_new_one_to_one_relationship`, and to avoid an extra query plus a behaviour change the PRD does not ask for. Happy to switch if you would rather have the 403.

packages/forest_admin_agent/spec/lib/forest_admin_agent/routes/resources/store_spec.rb
PMerlet: Fixed in 53da1410 — fixture is now `origin_type_value: 'Person::Legacy'` on a collection still named `person`, so the `'addressable_type' => 'person'` assertion only passes under `context.collection.name`. Verified: swapping the implementation to `schema.origin_type_value` now fails that test, where before it stayed green.

packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/store.rb
PMerlet: Pushing back on this one — I think the premise is off. Every one-to-one call site in `update_related.rb` already intersects the **foreign** collection's scope, not the parent's:

- `break_old_polymorphic_one_to_one_relationship` — `get_scope(context.child_collection)` (`:125`)
- `create_new_polymorphic_one_to_one_relationship` — `get_scope(context.child_collection)` (`:174`)
- `break_old_one_to_one_relationship` — `get_scope(context.child_collection)` (`:188`)
- `create_new_one_to_one_relationship` — `get_scope(context.child_collection)` (`:222`)

The only `get_scope(context.collection)` in that file is in `scoped_fk_owner_filter` (`:65`), and it feeds `context.collection.update` on the `ManyToOne` / `PolymorphicManyToOne` paths (`:78`, `:93`) — parent scope on a parent update, which is right.

The cited lines carry no `get_scope` call: `:104` is a `Collection.get_value`, `:153` a comment, `:167` a `match_ids`, `:201` the closing of a filter whose scope is the `:188` one.

So PRD-908 / #350 did use the foreign collection's scope on the Ruby side, the "mirroring `create_new_one_to_one_relationship`" claim in the description is accurate, and there is no parent-scope hole to follow up on. Description left as-is on this point, no follow-up opened. Shout if you are looking at something I am missing — happy to reopen.

hercemer42: Claude Opus 5 (claude-opus-5): Agreed on the outcome — nothing to follow up. But the premise wasn't off; we were reading different trees.

At `62274ca`, the base this review ran against, all four one-to-one sites did pass the parent's scope:

```
$ git show 62274ca:.../related/update_related.rb | grep -n "get_scope"
104:                  context.permissions.get_scope(context.collection),
153:                    context.permissions.get_scope(context.collection), new_fk_owner
167:                  context.permissions.get_scope(context.collection),
201:                  context.permissions.get_scope(context.collection),
```

No `scoped_fk_owner_filter` or `mutated_collection` in that revision either — `021deac` (#350, "scope and authorize related write routes on the mutated collection") is the commit that introduced both and switched those four to `context.child_collection`. It landed on main *after* this review was filed (main moved `62274ca` → `a1ad675`), and the rebase pulled it into this branch. So your line numbers are right now and mine were right then.

Net: the parent-scope hole was real, #350 closed it independently, "mirroring `create_new_one_to_one_relationship`" is accurate for the tree this will land on, and no follow-up is warranted. Resolving.

packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/store.rb
hercemer42: Claude Opus 5 (claude-opus-5): Contract stated and pinned by `store_spec.rb:355` — that settles it, resolving.

One stale detail: the PR body says the contract is "pinned by a test and stated at the call site", but `7a1a75c` dropped the comments on this path, so `link_one_to_one_relations` now carries none. Trim the clause or put the one line back — no code impact, the test is what holds the contract.

hercemer42 (approved):

PR #357, opened 2026-08-13
fix(agent): serialize every level of a projection deeper than one relation

PMerlet:
## Context

The `Forest-Projection` header ([#352](https://github.com/ForestAdmin/agent-ruby/pull/352)) accepts paths of any depth — `FieldValidator` recurses through relations — and the datasources fetch them. But the record routes built their JSON:API `include` list from `Projection#relations`, which only knows about the first hop. Everything below `author` was read from the database and then dropped at serialization time.

This is the behaviour agent-nodejs already has: its serializer walks registered relationships recursively, with no include list at all.

## What this fixes

- **`Projection#relation_include_paths`** returns the dotted path of every hop, so `author:company:name` yields `['author', 'author.company']`. Used by `show`, `list` and `list_related`.
- **`find_recursive_relationships` recursed with the parent's `class_name`**, so `author.company` was looked up in the parent schema and raised `InvalidIncludeError`. Depth 2 could not work even with the right include list.
- **A nested polymorphic hop now carries its type and foreign-key columns.** `add_polymorphic_type_fields` only covers the root level, so `author:documentable:*` arrived without `author:documentable_type` / `author:documentable_id` — and a target with neither a collection nor an id is silently dropped from the payload.
- **A bare nested polymorphic relation is expanded to `:*`** like a bare root one. `documentable` was normalised at the root but `author:documentable` returned a 400.

## Tests

Every fix has a test that fails without it.

- `projection_spec`: include paths for 1 / 3 hops, shared hops, polymorphic wildcard
- `forest_serializer_spec`: three-hop chain in `included` with its linkages, stop at the requested depth, absent intermediate record, nested polymorphic resolved from its type column, phantom target linkage rebuilt from the foreign key, unlinked target
- `query_string_parser_spec`: two-hop path kept, validation of the last field, type and key columns added, bare nested polymorphic normalised, field under a polymorphic rejected, to-many hop rejected at any depth
- `show_spec` / `list_spec` / `list_related_spec`: end to end on the three routes, including a related record shared by two rows appearing once in `included`
- `join_to_one_optimization_spec`: a two-hop path folded into one JOINed query, columns restricted per level, primary key of every level after `with_pks`, nil intermediate
- `active_record_serializer_spec`: nested polymorphic carries type and foreign key, plain field through a polymorphic one-to-one hop
- `n_plus_one_spec`: the preload path stays batched, whatever the number of rows

- `field_validator_spec`: the wildcard allowed, a field nested under the relation rejected, a field nested under the wildcard rejected

Suites run locally: `forest_admin_agent` (870), `forest_admin_datasource_toolkit` (472), `forest_admin_rails` (110), `forest_admin_datasource_active_record` (187), `forest_admin_datasource_customizer` (691), `forest_admin_datasource_rpc` (168). RuboCop clean.

## No new capability: consumers gate on `canUseProjectionViaHeader`

Considered and dropped, because this was a ruby-only divergence rather than a feature both agents are gaining.

agent-nodejs has honoured any depth since the day it shipped the header ([#1813](https://github.com/ForestAdmin/agent-nodejs/pull/1813), 2026-08-10) — that commit already tests `id,owner:address:street,owner:address:country:name`, and its get-one route serializes with no include list, so there was nothing to truncate. ruby shipped the same header in 1.38.0 (#352) but passed `relations(only_keys: true)` as the include list.

So `canUseProjectionViaHeader` has always meant "every hop is honoured" on node, and meant "truncated at one hop" only on ruby 1.38.0, 1.38.1 and 1.38.2 — three patches released between Aug 11 and Aug 13, superseded by the release carrying this PR. Gating depth ≥ 2 on the existing key is therefore correct for every node agent and every ruby ≥ 1.38.3, and it keeps a permanent public contract out of both agents to cover a window that closes before the frontend work (PRD-767 / ForestAdmin/forestadmin#9881) can consume it. A customer sitting on one of those three ruby patches needs a patch bump, not a handshake.

## Review follow-ups

- **`documentable:*:brand` used to reach the serializer.** `FieldValidator` destructured with a bare `split(':')`, so the suffix was `*`, the wildcard guard passed and the deeper-hop recursion was skipped. Harmless while include paths kept only the first hop; with every hop carried, the serializer looked for a `*` relationship and raised `JSONAPI::Serializer::InvalidIncludeError` — which descends from `Exception`, not `StandardError`, so the controller's rescue never ran and the client got a bare Rails 500 with nothing in the customer's logger. Now `split(':', 2)`, which makes it the 400 it always should have been, and the controller also rescues `JSONAPI::Serializer::Error`.
- The four new header-projection helpers are added to the existing `private_class_method` list instead of landing as public API on a released gem class, `POLYMORPHIC_TARGET_WILDCARD` is used in both places that build the wildcard, and the two blocks flagged as complex are named (`polymorphic_linkage_columns`, `field_along_path`).

## Known limits, unchanged by this PR

- An unknown value in a `*_type` column raises `Collection 'X' not found`. Verified identical at the root and nested — this PR makes the case reachable one hop further, it does not create it.
- Permissions and scopes are checked on the root collection only, at any depth, in this agent and in agent-nodejs. Traversing a relation through a projection is not permission-checked. Pre-existing, tracked separately.

### General

- [x] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org)
- [x] Test manually the implemented changes
- [x] Validate the code quality (indentation, syntax, style, simplicity, readability)

### Security

- [x] Consider the security impact of the changes made

> [!NOTE]
> ### Fix serialization of projections deeper than one relation in list, related-list, and show endpoints
> - Adds `Projection#relation_include_paths` to flatten nested relation projections into dot-separated paths (e.g. `author.company`), replacing the previous `relations(only_keys: true)` / `relations.keys` calls that only returned top-level relation names.
> - Updates list, related-list, and show route handlers to pass `relation_include_paths` to `JSONAPI::Serializer`, enabling correct compound-document includes for multi-hop relationships.
> - Refactors `ForestSerializerOverride#find_recursive_relationships` to resolve the serializer class and `class_name` once per path hop, fixing nested polymorphic to-one serialization.
> - Extends `QueryStringParser#build_header_projection_fields` with helpers (`expand_polymorphic_leaf`, `nested_polymorphic_linkage_fields`) to inject type/key linkage columns for nested polymorphic relations and normalize leaf polymorphic paths to wildcards.
> - Risk: serializer `include` payloads now contain intermediate dot-path segments that were previously omitted; any consumer relying on the old flat key list will see different include behavior.
>
> 
> #### Changes since #357 opened
>
> - Added require for `jsonapi-serializers` library in `ForestController` [6f82c52]
> - Added test coverage for `JSONAPI::Serializer` exception handling in `ForestController` [6f82c52]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 989c4ad.</sup>
> 
>

### DIFF
--- a/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/show.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/show.rb
@@ -38,7 +38,7 @@ module ForestAdminAgent
               class_name: context.collection.name,
               is_collection: false,
               serializer: Serializer::ForestSerializer,
-              include: projection.relations(only_keys: true)
+              include: projection.relation_include_paths
             )
           }
         end
--- a/packages/forest_admin_agent/lib/forest_admin_agent/utils/query_string_parser.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/utils/query_string_parser.rb
@@ -89,7 +89,7 @@ module ForestAdminAgent
           when 'Column'
             field_name
           when 'PolymorphicManyToOne'
-            "#{field_name}:*"
+            "#{field_name}:#{POLYMORPHIC_TARGET_WILDCARD}"
           else
             relation_fields = args.dig(:params, :fields, field_name)
 
@@ -136,10 +136,13 @@ module ForestAdminAgent
           each_field_along_path(collection, segments).flat_map do |field, index|
             next [] unless polymorphic_many_to_one?(field) && index.positive?
 
-            relation_path = segments[0...index]
-            [field.foreign_key_type_field, field.foreign_key].map { |column| (relation_path + [column]).join(':') }
+            polymorphic_linkage_columns(field, segments[0...index])
           end
         end
+      end
+
+      def self.polymorphic_linkage_columns(field, relation_path)
+        [field.foreign_key_type_field, field.foreign_key].map { |column| (relation_path + [column]).join(':') }
       end
 
       def self.each_field_along_path(collection, segments)
@@ -148,7 +151,7 @@ module ForestAdminAgent
         current_collection = collection
 
         segments.each_with_index do |segment, index|
-          field = index.zero? ? get_field(current_collection, segment) : current_collection.schema[:fields][segment]
+          field = field_along_path(current_collection, segment, root: index.zero?)
           break if field.nil?
 
           yield field, index
@@ -157,6 +160,10 @@ module ForestAdminAgent
 
           current_collection = collection.datasource.get_collection(field.foreign_collection)
         end
+      end
+
+      def self.field_along_path(collection, segment, root:)
+        root ? get_field(collection, segment) : collection.schema[:fields][segment]
       end
 
       def self.polymorphic_many_to_one?(field)
@@ -174,7 +181,10 @@ module ForestAdminAgent
               'Please check if the field name is correct.'
       end
       private_class_method :add_polymorphic_type_fields, :build_projection_fields,
-                           :build_header_projection_fields, :get_field
+                           :build_header_projection_fields, :get_field,
+                           :expand_polymorphic_leaf, :nested_polymorphic_linkage_fields,
+                           :polymorphic_linkage_columns, :each_field_along_path,
+                           :field_along_path, :polymorphic_many_to_one?
 
       def self.parse_projection_with_pks(collection, args)
         parse_projection_from_request(collection, args).with_pks(collection)
--- a/packages/forest_admin_agent/lib/forest_admin_agent/utils/query_string_parser.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/utils/query_string_parser.rb
@@ -9,6 +9,7 @@ module ForestAdminAgent
 
       DEFAULT_ITEMS_PER_PAGE = '15'.freeze
       DEFAULT_PAGE_TO_SKIP = '1'.freeze
+      POLYMORPHIC_TARGET_WILDCARD = '*'.freeze
 
       def self.parse_condition_tree(collection, args)
         filters = begin
@@ -110,18 +111,56 @@ module ForestAdminAgent
         root_field_names_with_types = root_field_names.dup
         add_polymorphic_type_fields(collection, root_field_names_with_types)
 
-        projection_fields = requested_paths.map do |path|
-          field_name, nested_path = path.split(':', 2)
-          field = get_field(collection, field_name)
+        projection_fields = requested_paths.map { |path| expand_polymorphic_leaf(collection, path) }
 
-          if field.type == 'PolymorphicManyToOne' && (nested_path.nil? || nested_path == '*')
-            "#{field_name}:*"
-          else
-            path
-          end
+        projection_fields |
+          (root_field_names_with_types - root_field_names) |
+          nested_polymorphic_linkage_fields(collection, projection_fields)
+      end
+
+      def self.expand_polymorphic_leaf(collection, path)
+        segments = path.split(':')
+        leaf_index = segments.size - 1
+
+        each_field_along_path(collection, segments) do |field, index|
+          return "#{path}:#{POLYMORPHIC_TARGET_WILDCARD}" if polymorphic_many_to_one?(field) && index == leaf_index
         end
 
-        projection_fields | (root_field_names_with_types - root_field_names)
+        path
+      end
+
+      def self.nested_polymorphic_linkage_fields(collection, projection_fields)
+        projection_fields.flat_map do |path|
+          segments = path.split(':')
+
+          each_field_along_path(collection, segments).flat_map do |field, index|
+            next [] unless polymorphic_many_to_one?(field) && index.positive?
+
+            relation_path = segments[0...index]
+            [field.foreign_key_type_field, field.foreign_key].map { |column| (relation_path + [column]).join(':') }
+          end
+        end
+      end
+
+      def self.each_field_along_path(collection, segments)
+        return to_enum(:each_field_along_path, collection, segments) unless block_given?
+
+        current_collection = collection
+
+        segments.each_with_index do |segment, index|
+          field = index.zero? ? get_field(current_collection, segment) : current_collection.schema[:fields][segment]
+          break if field.nil?
+
+          yield field, index
+
+          break unless field.respond_to?(:foreign_collection)
+
+          current_collection = collection.datasource.get_collection(field.foreign_collection)
+        end
+      end
+
+      def self.polymorphic_many_to_one?(field)
+        field.type == 'PolymorphicManyToOne'
       end
 
       def self.get_field(collection, field_name)
--- a/packages/forest_admin_datasource_toolkit/lib/forest_admin_datasource_toolkit/components/query/projection.rb
+++ b/packages/forest_admin_datasource_toolkit/lib/forest_admin_datasource_toolkit/components/query/projection.rb
@@ -41,6 +41,12 @@ module ForestAdminDatasourceToolkit
           only_keys ? relations.keys : relations
         end
 
+        def relation_include_paths
+          relations.flat_map do |relation, sub_projection|
+            [relation, *sub_projection.relation_include_paths.map { |path| "#{relation}.#{path}" }]
+          end
+        end
+
         def nest(prefix: nil)
           prefix ? Projection.new(map { |path| "#{prefix}:#{path}" }) : self
         end
--- a/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/list.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/list.rb
@@ -43,7 +43,7 @@ module ForestAdminAgent
               class_name: context.collection.name,
               is_collection: true,
               serializer: Serializer::ForestSerializer,
-              include: projection.relations(only_keys: true),
+              include: projection.relation_include_paths,
               meta: handle_search_decorator(args[:params]['search'], records, context.collection)
             )
           }
--- a/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/related/list_related.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/related/list_related.rb
@@ -53,7 +53,7 @@ module ForestAdminAgent
                 is_collection: true,
                 class_name: context.child_collection.name,
                 serializer: Serializer::ForestSerializer,
-                include: projection.relations.keys
+                include: projection.relation_include_paths
               )
             }
           end
--- a/packages/forest_admin_agent/lib/forest_admin_agent/serializer/forest_serializer_override.rb
+++ b/packages/forest_admin_agent/lib/forest_admin_agent/serializer/forest_serializer_override.rb
@@ -54,23 +54,24 @@ module ForestAdminAgent
               # Full linkage: a request for comments.author MUST automatically include comments
               # in the response.
               objects = is_collection ? object : [object]
+
+              relation = ForestAdminAgent::Facades::Container.datasource
+                                                             .get_collection(options[:class_name].gsub('::', '__'))
+                                                             .schema[:fields][attribute_name]
+              if relation.type == 'PolymorphicManyToOne'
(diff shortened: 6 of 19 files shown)


packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/show.rb
```diff
@@ -38,7 +38,7 @@ def handle_request(args = {})
               class_name: context.collection.name,
               is_collection: false,
               serializer: Serializer::ForestSerializer,
-              include: projection.relations(only_keys: true)
+              include: projection.relation_include_paths
```
hercemer42, line 41: Claude Fable 5 (claude-fable-5): **Must fix**

**Applies to:** `packages/forest_admin_agent/lib/forest_admin_agent/routes/capabilities/collections.rb:67-70` (not in this diff). Anchored here because this is the get-one route the capability gates.

This fix needs a new capability key, or the frontend cannot consume it and three released versions keep silently losing data.

`canUseProjectionViaHeader: true` has been announced since v1.38.0 (#352, `ff25d09`). It is announced identically by v1.38.0, v1.38.1, v1.38.2 — none of which have this fix — and by whatever release this PR lands in. So the frontend has **no runtime way to tell a patched agent from an unpatched one**, and `canUseProjectionViaHeader` currently means "I accept the header", not "I honour every hop of it".

What a customer on 1.38.2 sees today, and will keep seeing after this merges: a workspace chart templated on `{{currentRecord.owner.address.street}}` sends `owner:address:street`, the agent validates it, the datasource fetches it, the serializer drops everything below `owner`, and the page renders an empty value. HTTP 200, no error payload, nothing in the logger. The frontend work that will emit these paths (ForestAdmin/forestadmin#9881, PRD-767) has to gate on *something*, and right now the only options are a minimum-version check on the agent — which the capabilities handshake exists to avoid — or not shipping depth ≥2 at all.

Suggested: add `canUseDeepProjectionViaHeader: true` alongside the existing three in this release (`canUseNestedRelationProjection` if you prefer the `canUseMultipleFieldsProjectionOnRelation` shape). Same key in agent-nodejs, where it can be `true` unconditionally — its serializer takes no include list and never had this bug, so the flag is purely the signal ruby needs.

Worth pairing with a line in the PR description: consumers gate on the new key, not on `canUseProjectionViaHeader`.

packages/forest_admin_datasource_toolkit/lib/forest_admin_datasource_toolkit/components/query/projection.rb
```diff
@@ -41,6 +41,12 @@ def relations(only_keys: false)
           only_keys ? relations.keys : relations
         end
 
+        def relation_include_paths
```
hercemer42, line 44: Claude Fable 5 (claude-fable-5): **Should fix**

**Applies to:** the real fix belongs in `packages/forest_admin_datasource_toolkit/lib/forest_admin_datasource_toolkit/validations/field_validator.rb:25` (not in this diff). This line is the anchor because it is what turns the bad path into an extra include hop.

A client sending `Forest-Projection: <polymorphic_relation>:*:anything` now gets an **unlogged Rails 500** from `show` / `list` / `list_related`, where `main` returned 200. No Forest `{errors: [...]}` payload reaches the client and nothing reaches the customer's configured logger, so the failure is invisible in their logs.

The chain:

- `FieldValidator.validate` destructures with a bare `split(':')` (`field_validator.rb:25`), so `documentable:*:brand` gives `prefix="documentable"`, `suffix="*"`. The `suffix != '*'` guard at `:30` therefore passes, and the polymorphic branch skips the recursive call at `:40`. The path survives validation untouched — `expand_polymorphic_leaf` leaves it alone too, since the polymorphic hop is not the leaf.
- `relation_include_paths` flattens it to `["documentable", "documentable.*"]`. The old `relations(only_keys: true)` gave `["documentable"]`, which is why this was harmless before.
- On a row where `documentable_type` and `documentable_id` are both set, the recursion reaches the `"*"` level, finds no relationship by that name, and raises `JSONAPI::Serializer::InvalidIncludeError` (`forest_serializer_override.rb:40`).
- `InvalidIncludeError < JSONAPI::Serializer::Error < Exception` — **not** `StandardError` (jsonapi-serializers 1.0.1, `lib/jsonapi-serializers.rb:7-9`; the gemspec pins `~> 1.0`). `forest_controller.rb:25` rescues only `StandardError`, so `exception_handler` never runs and the `Container.logger.log('Error', …)` call at `forest_controller.rb:104` never fires. The exception unwinds to Rails' default 500.

It is data-dependent: an unlinked row short-circuits at `forest_serializer_override.rb:52` and still returns 200, so the same request succeeds or 500s depending on the record.

`field.split(':', 2)` at `field_validator.rb:25` is the cheap fix — the wildcard guard then sees the whole remainder and `documentable:*:brand` becomes the 400 it always should have been. Worth pairing with a `rescue JSONAPI::Serializer::Error` at the routes, because every other error in that hierarchy escapes the same way.

packages/forest_admin_agent/lib/forest_admin_agent/utils/query_string_parser.rb
```diff
@@ -110,18 +111,56 @@ def self.build_header_projection_fields(collection, requested_paths)
...
-          if field.type == 'PolymorphicManyToOne' && (nested_path.nil? || nested_path == '*')
-            "#{field_name}:*"
-          else
-            path
+        projection_fields |
+          (root_field_names_with_types - root_field_names) |
+          nested_polymorphic_linkage_fields(collection, projection_fields)
+      end
+
+      def self.expand_polymorphic_leaf(collection, path)
```
hercemer42, line 121: Claude Fable 5 (claude-fable-5): **Preferential**

**Applies to:** the `private_class_method` call at `query_string_parser.rb:176` (not in this diff).

`expand_polymorphic_leaf`, `nested_polymorphic_linkage_fields`, `each_field_along_path` and `polymorphic_many_to_one?` are all internal to header parsing, but none of them was added to the `private_class_method` list that already covers `add_polymorphic_type_fields`, `build_projection_fields`, `build_header_projection_fields` and `get_field`. They land as public API on a released gem class, so renaming or reshaping any of them later is a breaking change for anyone who found them.

Adding all four to the existing call is enough — `to_enum(:each_field_along_path, …)` and the implicit-receiver `get_field` call both still resolve once they are private.

packages/forest_admin_agent/lib/forest_admin_agent/utils/query_string_parser.rb
```diff
@@ -9,6 +9,7 @@ class QueryStringParser
 
       DEFAULT_ITEMS_PER_PAGE = '15'.freeze
       DEFAULT_PAGE_TO_SKIP = '1'.freeze
+      POLYMORPHIC_TARGET_WILDCARD = '*'.freeze
```
hercemer42, line 12: Claude Fable 5 (claude-fable-5): **Code style**

**Applies to:** `query_string_parser.rb:92` (not in this diff).

The new constant replaces the literal at `:126`, but `build_projection_fields` still builds `"#{field_name}:*"` at `:92`, so someone grepping for `POLYMORPHIC_TARGET_WILDCARD` finds one of the two places that produce the wildcard. Either use it in both, or drop it and leave `'*'` inline at `:126`.

hercemer42 (changes requested):

packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/show.rb
PMerlet: Not doing this one — the signal already exists, because this was a ruby-only divergence rather than a capability both agents are gaining.

agent-nodejs has honoured any depth since the day it shipped the header (#1813, 2026-08-10): that commit already tests `id,owner:address:street,owner:address:country:name`, and `get.ts` serializes with no include list, so there was nothing to truncate. ruby shipped the same header in 1.38.0 (#352) but passed `relations(only_keys: true)`.

So `canUseProjectionViaHeader` has always meant "every hop is honoured" on node, and meant "truncated at one hop" only on ruby 1.38.0 / 1.38.1 / 1.38.2 — three patches released between Aug 11 and Aug 13, superseded by the release that carries this PR. Gating depth >= 2 on the existing key is correct for every node agent and every ruby >= 1.38.3.

On the premise: the frontend cannot emit `owner:address:street` today, so no customer on 1.38.2 is losing data through it. `serializeProjectionAsHeader` builds the header from a flat `fields` map and expands exactly one level (`${fieldName}:${subField}`), and all three call sites (`basic-adapter`, `relationship-store`, `csv-exporter`) go through it. The paths appear only once the frontend's projection structure grows a hop (PRD-767 / ForestAdmin/forestadmin#9881) — by which time these three ruby patches are behind us.

Adding the key on ruby alone would also have been the riskier move: node does not advertise it and never had the bug, so a frontend gating on it would downgrade node agents that already support depth >= 2.

Decision is Pierre's; reasoning recorded in the PR description so it does not have to be re-derived. Leaving this thread open for you to close or push back.

packages/forest_admin_datasource_toolkit/lib/forest_admin_datasource_toolkit/components/query/projection.rb
PMerlet: Fixed in 066c1d51. Confirmed the whole chain, including `class Error < Exception` in the installed jsonapi-serializers 1.0.1 and the `rescue StandardError` at `forest_controller.rb:25` — the exception did escape unlogged.

`field_validator.rb:25` now destructures with `split(':', 2)`, so the wildcard guard sees the whole remainder and `documentable:*:brand` is the 400 it always should have been. That also made the recursion's `field[dot_index + 1, ...]` recomputation redundant, so it now reuses the suffix it already had.

Took the pairing suggestion too: the controller rescues `JSONAPI::Serializer::Error` alongside `StandardError`, since every error in that hierarchy escapes the same way.

Three tests in `field_validator_spec`: the wildcard allowed, a field nested under the relation rejected, a field nested under the wildcard rejected. Reverting to `split(':')` fails exactly the last one.

packages/forest_admin_agent/lib/forest_admin_agent/utils/query_string_parser.rb
PMerlet: Done in 076e32cc — the four helpers plus the two extracted since (`polymorphic_linkage_columns`, `field_along_path`) are on the existing `private_class_method` call. Checked there was no caller outside the class first, and `to_enum` and the implicit-receiver `get_field` both still resolve.

packages/forest_admin_agent/lib/forest_admin_agent/utils/query_string_parser.rb
PMerlet: Done in 076e32cc — `build_projection_fields` uses `POLYMORPHIC_TARGET_WILDCARD` too, so both places that produce the wildcard are greppable.

packages/forest_admin_agent/lib/forest_admin_agent/routes/resources/show.rb
hercemer42: Accepted, and the ruby-only-key argument is the decisive one — node never advertised a separate key and never had the bug, so a frontend gating on one would have downgraded agents that already honour every hop. That cost is real and immediate; the thing it buys is not.

Verified your premise independently rather than taking it on trust: `serializeProjectionAsHeader` expands exactly one level out of a flat `fields` map, and the only builder feeding it, `buildRecordProjection` (`crud-handler.ts`), returns `Map<string, Set<string>>` of plain sub-field names. No call site can produce `owner:address:street` today, so nobody on 1.38.x is losing data through it. My comment overstated that as present-tense impact.

One residual, for PRD-767 rather than for this PR: "by which time these three ruby patches are behind us" holds for the release timeline, not for deployed agents — customers pin, and 1.38.0/1/2 will go on announcing `canUseProjectionViaHeader: true` while truncating at one hop. If ForestAdmin/forestadmin#9881 wants belt-and-braces it can floor ruby at 1.38.3, which is a version check rather than a capability, and costs node nothing. Narrow band, but the failure mode is silent. Recording it on the ticket; not blocking here.

Resolving.

hercemer42 (approved):

